PR #19, opened 2016-08-17
Ropes save image link

### LINKED ISSUE #2: Save Image Link
#### As a user I want to save a link to an image.

__Story__:

Now that we have an awesome landing page for our users to experience it is time
to introduce our first feature! In general, we want to deliver value to our
customers as fast as possible. As a result, we have decided that we want our
users to be able to save links to images, and immediately see that we have
saved the URL to the image.

One could argue that simply saving the image, and not showing it, provides some
value. Yes, you could split this story here. However, because saving image URLs
and viewing those images are both relatively small additions that work well
together, it is acceptable to complete them together. Note, on the other hand,
that we do not yet want to implement the image index. Its addition adds too
much complexity, so we’ll save that for a future story.

In general, feature addition, or stories, should be split into the smallest
deliverable chunk that provides value to your users with the caveat that it
might make sense to combine two, sometimes more, tiny and related features into
a single story.

__Acceptance criteria__:
- [x] The landing page links to the image link submission form.
- [x] The link is entered through a form.
- [x] After the form is saved, the link is persisted in the database.
- [x] After the form is submitted, the user is redirected to a page displaying
  the image.

__Dependencies__:
- Deployment Environment

__Note__: Many image sharing sites will permit one to upload images, or
download images when provided a link. For this project we will only work with
the link, i.e., URL, to various images.

kkajla12:
Closes #2


app/controllers/images_controller.rb
```diff
@@ -0,0 +1,14 @@
+class ImagesController < ApplicationController
+  def new
+    @image = Image.new
+  end
+
+  def create
+    @image = Image.create(url: params['image']['url'])
```
HaominZhang: It would be better if we use [strong parameter](http://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters) here.

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+    redirect_to image_path(id: @image.id)
```
HaominZhang: Is it possible that we save this url fail? Do you want to add some error handling?

db/migrate/20160817172711_add_image.rb
```diff
@@ -0,0 +1,7 @@
+class AddImage < ActiveRecord::Migration[5.0]
+  def change
+    create_table :images do |t|
+      t.string :url
```
HaominZhang: Do we allow saving an empty url? If not, we could add `null: false` to prevent it from database side

db/schema.rb
```diff
@@ -10,6 +10,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 0) do
+ActiveRecord::Schema.define(version: 20160817172711) do
+
+  create_table "images", force: :cascade do |t|
+    t.string "url"
```
HaominZhang: Same as https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/19/commits/520f6ea0249decbd78421d9a1b8b7b2d5443588c#r75228214

app/models/image.rb
```diff
@@ -0,0 +1,2 @@
+class Image < ActiveRecord::Base
```
HaominZhang: If we don't allow empty url in our case, besides handle it from database side, we could also add some validations in model to do that. How do you think?

app/controllers/images_controller.rb
```diff
@@ -5,7 +5,7 @@ def new
 
   def create
     @image = Image.create(url: params['image']['url'])
-    redirect_to image_path(id: @image.id)
+    redirect_to image_path(@image)
```
HaominZhang: How about moving this change out from this commit since this commit message says `add images controller tests`?

HaominZhang:
I think you did a good job on creating another commit for testing. And one potential improvements could be make the commit more atomic. For example, we can have a commit containing adding the model, migration and unit test to this model. And we can have another commit that only contains adding the `new` method in controller and its controller test, and so on. How do you think about it?

app/controllers/images_controller.rb
kkajla12: resolved in 41f1c710b62604605a8e23f5639cccd2734ec73d

app/controllers/images_controller.rb
kkajla12: resolved in 41f1c710b62604605a8e23f5639cccd2734ec73d

db/migrate/20160817172711_add_image.rb
kkajla12: resolved in ec0f52a3f03848b053c0b915db12fe144e77f692

db/schema.rb
kkajla12: resolved in ec0f52a3f03848b053c0b915db12fe144e77f692

app/models/image.rb
kkajla12: resolved in ec0f52a3f03848b053c0b915db12fe144e77f692

app/controllers/images_controller.rb
kkajla12: Since I redid all of the commits for this issue, this is no longer a problem. I will take note of how to handle syntax changes like this in the future. Thank you.

kkajla12:
I have submitted all of the changes you requested. I did not split up addition of the create and show actions since they are somewhat dependent on each other, so I think that is as small as I can reasonably make that commit. Please look over the code again and let me know what you think. Thanks.

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,4 @@
+require 'test_helper'
```
HaominZhang: Seems you have the test case in next commit, do you mind moving the creation of this file into the next one? And we can have an atomic commit that only contains changes about data (model and migration). How do you think?

test/controllers/images_controller_test.rb
```diff
@@ -1,9 +1,28 @@
...
+    assert_select 'img', 1
+  end
+
+  test 'should create image' do
+    post images_path('image' => {
+      'url' => 'http://pets.petsmart.com/services/_images/grooming/dog/m_t/dog-aromatherapy.jpg'
+    })
+    assert_response :found
+    assert_not_nil Image.find_by(url: @url)
+  end
```
HaominZhang: How about adding another case to test you get the error message if saving failed?

HaominZhang:
@kkajla12 I left some specific comment in your code, please take a look. And I also have some general suggestion at this moment
1. From the reviewer perspective, it would be a better practice when you address some comments, don't squash into existing commits. [Modifying code after initial code review](https://sites.google.com/a/appfolio.com/eng/general-engineering-resources/code-review)
2. Regarding the atomicity, I totally agree that the `create` and `show` are dependent from usage perspective. But let's think about from functionality side. For `create`, it takes the url you submitted through the form and save it into database. And for `show`, it fetches the url from database and shows the corresponding image. These two actions are independent to this degree. And commit don't have to be usage deliverable, so we don't need to worry about we must display the image after we save the url. So do you think it make sense to separate it into two more atomic commits?
3. Seems we are missing some unit tests (success and failure), could you add them as well?

HaominZhang:
@kkajla12 Everything looks good to me right now. A minor suggestion is that do you mind squashing the last several commits about adding test and return 404 response into its corresponding original commit? Actually we don't really need to do this, but it would be better for this simple story. How do you think?

HaominZhang:
And also, can we add the flow test as well?

HaominZhang:
LGTM! Feel free to merge it.

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,21 @@
+class ImagesController < ApplicationController
+  def new
+    @image = Image.new
+  end
+
+  def create
+    params.require(:image).permit(:url)
```
bboe: This line isn't necessary if you are going to pull out the individual field in the call to `create`.

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    @image = Image.create(url: params['image']['url'])
+    if @image.errors.any?
```
bboe: Does this work like you think it does? I didn't see any tests for it.

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      render :new
+    else
+      redirect_to image_path(@image)
```
bboe: Is there a message you could add to let the user know that things worked as intended? I realize the image itself probably would do that, but when you're doing multiple things it's easy to lose track of how you got somewhere.

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+  end
+
+  def show
+    @image = Image.find(params['id'])
+  rescue ActiveRecord::RecordNotFound => _e
+    render plain: 'Oops! The image you are looking for was not found!', status: :not_found
```
bboe: I like that you are using a custom 404 response here. Just FYI, rescuing this exception isn't strictly required. Rails provides a standard 404 page when an `ActiveRecord::RecordNotFound` exception occurs.

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    if @image.errors.any?
+      render :new
+    else
+      redirect_to image_path(@image)
+    end
+  end
+
+  def show
+    @image = Image.find(params['id'])
+  rescue ActiveRecord::RecordNotFound => _e
```
bboe: ` => _e` is that necessary?

db/migrate/20160817172711_add_image.rb
```diff
@@ -0,0 +1,7 @@
+class AddImage < ActiveRecord::Migration[5.0]
+  def change
+    create_table :images do |t|
+      t.string :url, null: false
```
bboe: Whenever you add a new table please add the timestamps too unless you really don't want them. Perhaps you've added that later, but it's a good idea to get in the habit of doing it.

cc @HaominZhang

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    get image_path(id: -1)
+    assert_response :not_found
+    assert_equal 'Oops! The image you are looking for was not found!', response.body
+  end
+
+  test 'should create image' do
+    post images_path('image' => {
+      'url' => 'http://pets.petsmart.com/services/_images/grooming/dog/m_t/dog-aromatherapy.jpg'
+    })
+    assert_response :found
```
bboe: Try using `assert_redirected_to`. That encompasses `assert_response` and you can check to ensure you are being redirected to the right path.

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    assert_not_nil Image.find_by(url: @url)
+  end
+
+  test 'should not allow empty url' do
+    post images_path('image' => {
+      'url' => ''
+    })
+    assert_response :ok
```
bboe: This sounds like a failure case. Should the response be `ok`?

test/flows/image_flow_test.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    page.find('#image_url').set(@url)
+    page.click_button('Upload URL')
+
+    assert has_current_path? image_path(1)
+    assert_equal page.find('h1').text, 'Your Image'
+    assert page.has_css?("img[src='#{@url}']")
+  end
+
+  def test_create_null_image
```
bboe: Do these two tests add any value over the controller tests?

Think of a flow as what a user would actually do to. In this case the flow might be:

* visit the home page
* click on new image
* make a mistake filling in the form
* successfully submit the form
* observe their new creation in all its glory

test/models/image_test.rb
```diff
@@ -0,0 +1,16 @@
+require 'test_helper'
+
+class ImageTest < ActiveSupport::TestCase
+  def test_url__valid
+    image = Image.new(url: 'https://test.com')
+
+    assert image.valid?
```
bboe: While not immediately obvious, we want to avoid using refute and assert directly. If you need to use them, I always recommend adding a third argument which is the message to be displayed when the assertion failed. In this particular case you are calling a method that returns a boolean and takes no arguments. The refute_predicate and assert_predicate functions are perfect for those cases. Please update in your code where appropriate:

assert_predicate image, :invalid?
Note I usually try to prefer assert over refute if a method exists that doesn't require a negation operator.

db/migrate/20160817172711_add_image.rb
HaominZhang: @bboe Thanks for addressing this. A question from me is how about `created_at` and `updated_at`? Do we also want to add them at this point of time too?

bboe: @HaominZhang yes, they should be added at this time via `t.timestamps` as shown here: http://edgeguides.rubyonrails.org/active_record_migrations.html

app/controllers/images_controller.rb
kkajla12: I knew about that, but I wanted to learn how to add my own and also just add a more relevant custom message.

db/migrate/20160817172711_add_image.rb
kkajla12: Noted. I made the mistake of not adding them here, but I did add them in the following migration.

test/models/image_test.rb
kkajla12: @bboe, it would be `assert_predicate image, :valid?` in this situation, right?

test/controllers/images_controller_test.rb
kkajla12: Actually, I'm not sure now. In the controller, if creation of the image record fails model validation, the controller re-renders the new image page. I was thinking I could check for a redirect instead, but that isn't actually a redirect, so how can I check that the new image page is re-rendered?

bboe: When a failure occurs with a form we typically render a 422 `:unprocessable_entity`, not a 200 ok. That's what I was asking about.

PR #21, opened 2016-08-23
Ropes image index

### LINKED ISSUE #3: Image Index
#### As a user I want the homepage to display all saved images.

__Story__:

Now that we are saving images for our users, we want them to be able to view
the list of images in the system. You might be thinking, "what if my users only
want to view images of cats", or "what if my users only want to see the images
they uploaded". Those are great questions. Maybe you will have the opportunity
to add those features in a future story. For now, we will keep it simple and
quickly deliver a little more value to our users.

__Acceptance criteria__:
- [x] New images that are added show up on the homepage.
- [x] These images are persisted if the browser is closed or even if the
  server is restarted.
- [x] Images are not displayed wider than 400px.
- [x] Newest images appear first.

__Dependencies__:
- Save Image Link

kkajla12:
Closes #3


app/assets/stylesheets/images.css
```diff
@@ -0,0 +1,3 @@
+.image-display {
+    max-width: 400px;
```
aaronpanch: This looks like more than two spaces. Double check your indentation settings.

app/controllers/images_controller.rb
```diff
@@ -1,4 +1,8 @@
 class ImagesController < ApplicationController
+  def index
+    @images = Image.order(created_at: :desc).all
```
aaronpanch: You don't need the `all` at the end here.

app/views/images/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>Images</h1>
+<% @images.each do |image| %>
+  <div>
```
aaronpanch: Having a `div` without any classes is sort of a "bad smell" -- you're probably using it to make sure each image appears on it's own line, and that's fine for now. In subsequent stories this should be refactored.

test/controllers/images_controller_test.rb
```diff
@@ -3,7 +3,24 @@
 class ImagesControllerTest < ActionDispatch::IntegrationTest
   setup do
     @url = 'https://upload.wikimedia.org/wikipedia/commons/2/26/YellowLabradorLooking_new.jpg'
-    @image = Image.create(url: @url)
+    @image = Image.create!(url: @url)
+  end
+
+  test 'should get index' do
+    @image.update_attributes!(url: 'http://images/1', created_at: Time.zone.now)
```
aaronpanch: Could you explain your thinking behind this line, and its role in this test?

app/views/images/index.html.erb
kkajla12: noted

app/assets/stylesheets/images.css
kkajla12: I checked my .editoconfig file and it sets the indent to 2 spaces, but for some reason the CSS files are adding 4 spaces. I checked this on Ryan's RubyMine too, and it's the same behavior. Other than the .editorconfig file, I'm not sure where else I can change this setting from. I ran through all of the settings and wasn't able to find anything related.

test/controllers/images_controller_test.rb
kkajla12: I paired with Jim Herzberg for this issue, and I think our reasoning was: because I didn't surround the create statement with error handling code, it was better to use the bang since that version of the function will at least throw an exception and generate a useful/relevant error. The version of the function without the bang returns a boolean which should then be explicitly handled in order to produce a proper error.

app/controllers/images_controller.rb
kkajla12: Resolved in 62581c3f467f04d5bcfb0ec29a27d11834791cb4

test/controllers/images_controller_test.rb
kkajla12: Although, I think you're probably asking about why I didn't just create the image inside of the first test like I did for the other tests to keep consistency. I don't have an answer for that, and I should probably change it to stay consistent with the other tests. I think when I first wrote the tests, I was following the style of having global objects, but it made more sense to do it the other way for subsequent tests and I didn't change the existing test.

db/migrate/20160819171839_add_timestamps_to_images.rb
```diff
@@ -0,0 +1,10 @@
+class AddTimestampsToImages < ActiveRecord::Migration[5.0]
+  def change
+    add_timestamps(:images, null: true)
```
bboe: Great job including `null: true`.

db/migrate/20160819171839_add_timestamps_to_images.rb
```diff
@@ -0,0 +1,10 @@
...
+    update <<-SQL
+      UPDATE images set created_at = '#{Time.zone.now.to_s(:db)}', updated_at = '#{Time.zone.now.to_s(:db)}'
+    SQL
+    change_column_null(:images, :created_at, false)
+    change_column_null(:images, :updated_at, false)
+  end
+end
```
bboe: Did you verify that this migration was reversible?

test/controllers/images_controller_test.rb
```diff
@@ -1,9 +1,21 @@
 require 'test_helper'
 
 class ImagesControllerTest < ActionDispatch::IntegrationTest
-  setup do
-    @url = 'https://upload.wikimedia.org/wikipedia/commons/2/26/YellowLabradorLooking_new.jpg'
-    @image = Image.create(url: @url)
+  test 'should get index' do
```
bboe: Does the index route exist at this point in the commit history? It seems like it's added in the following commit.

test/controllers/home_controller_test.rb
```diff
@@ -1,9 +0,0 @@
-require 'test_helper'
-
-class HomeControllerTest < ActionDispatch::IntegrationTest
-  test 'should get index' do
-    get root_path
-    assert_response :ok
-    assert_select 'p', 'Welcome to Image Sharer!'
-  end
-end
```
bboe: Nice atomic commit.

db/migrate/20160819171839_add_timestamps_to_images.rb
kkajla12: Yeah, I worked on this one with Jim and we did a migrate followed by a rollback and it worked.

bboe: Awesome! Good work.

PR #22, opened 2016-08-23
use validates_url gem to validate image urls when created

### LINKED ISSUE #4: Image URL Validation
#### As a user I want to be prevented from submitting invalid image URLs.

__Acceptance criteria__:
- [x] I cannot successfully save an image with an invalid URL.
- [x] An error message is associated with the appropriate input field on
  failure.

__Discussion Topic__:
- [x] What is a valid image URL?

__Dependencies__:
- Save Image Link

kkajla12:
Closes #4


app/models/image.rb
```diff
@@ -1,3 +1,4 @@
 class Image < ActiveRecord::Base
   validates :url, presence: true
+  validates :url, url: { allow_blank: true }
```
XanderStrike: If you change this line to `allow_blank: false` you don't need the presence validation.

XanderStrike: Also, does this gem have any way to ensure that the URL ends with an image extension? You could have display issues if someone submits `http://example.com` and your code tries to display that as an image.

kkajla12: So, I paired with Stanley Fung to implement this issue, and our reasoning behind adding the ```allow_blank: true``` and keeping the presence validation was to provide better/more specific error messages. With the proposed change, the error message produced for a blank URL is

> is not a valid URL

and keeping it the way we have it would produce

> can't be blank

I think it would be more useful to leave it as is, but having explained our reasoning, I would like to hear your next thoughts.

kkajla12: Regarding the second comment: as far as I know, this gem does not validate any specific file extensions. However, Stanley and I did some research and found that it's not necessarily true that all image URLs will end with the appropriate image file extension (I have some examples in my seed file). I could definitely add a validation function in addition to the gem to ensure the proper file extension if you recommend it.

XanderStrike: Technically empty string isn't a valid URL 😉

I appreciate that you've thought through all my questions and have good reasons for why you made the decisions you did. I think it's fine to leave it as-is.

XanderStrike:
👍

app/models/image.rb
kkajla12: Haha, that's definitely true. It's probably not a big deal to have a separate, specific error message for empty URLs, but I thought it would be good practice to provide nice, descriptive error messages for different scenarios. Thanks for the review 😄

test/models/image_test.rb
```diff
@@ -13,4 +13,11 @@ def test_url__must_be_present
     refute image.valid?
     assert_equal "can't be blank", image.errors.messages[:url].first
   end
+
+  def test_url__invalid
+    image = Image.new(url: 'asdf')
+
+    refute image.valid?
```
bboe: Please see my other comment regarding use of `assert_predicate`.

app/models/image.rb
bboe: Just FYI these two lines can be combined into:

    validates :url, presence: true, url: { allow_blank: true }

PR #23, opened 2016-08-24
update seeds.rb to populate database with 20 images upon setup

### LINKED ISSUE #5: Image Seeds
#### As a developer I want at least 20 images to be created when I setup the database.

__Story__:

Working on this application has been a lot of fun, right? You are fast
becoming a git expert, and an expert at working with a deployment
pipeline. Speaking of development and deployment pipelines, is it not pretty
annoying that everytime a new deployment is created it starts out with an empty
database? It seems like everytime a review application is created we need to
manually add a handful of images in order to manually test the
application. Would it not be great if each deployment was created with a
handful of existing images?

__Acceptance criteria__:
- [ ] After running `rake db:setup` at least 20 unique images are shown in the
  homepage.
- [ ] When a new app is deployed on heroku (e.g., every review app), that app
  automatically has at least 20 unique images on its homepage.

__Dependencies__:
- Image Index

kkajla12:
Closes #5


db/seeds.rb
```diff
@@ -5,3 +5,15 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
+
+(1..4).each do
```
mlewisno: So I saw that Ryan did this to, but it would be great if there were 20 separate images - it makes me checking that there really are 20 images a lot easier. Also, while your use of the fact that `create` can take in an array of hashes is correct, its a bit weird to have that while also using the `(1..4).each do` block.

Also, I'd like you to consider what would happen if one of your url strings wasn't a properly formatted string and the Image failed to save. With the code as it is now - would you know? How would you know?

On a side note, 4.times is much more readable and therefor preferable to `(1..4).each`

One last "bonus challenge" if you want ;) See if you can start with an array of strings and turn it into a the hash that you need to pass into `create!`

Also, unless you need the commented code above for a future thing (and even then) you should remove it. It just bloats the file and makes a read-through more difficult :)

mlewisno:
LGTM!

PR #24, opened 2016-08-26
Ropes save image link additions

kkajla12:


bboe:
Adding this file should resolve that issue: https://github.com/AppFolioOnboarding/base/blob/a2f567bc1b132532d3b894f3947a8aedc4d956b8/test/.rubocop.yml

app/controllers/images_controller.rb
```diff
@@ -13,6 +13,7 @@ def create
     if @image.errors.any?
       render :new
     else
+      flash.notice = 'Image created successfully!'
```
bboe: Now that we have a side-effect, should we include testing for it in our controller tests?

app/views/images/index.html.erb
```diff
@@ -1,4 +1,5 @@
 <h1>Images</h1>
+<%= link_to('Create Image', new_image_path) %>
```
bboe: Now that this exists, can you start all of your flow tests from the root path?

app/controllers/images_controller.rb
kkajla12: Yeah, that's my next order of business. I wanted to have one branch for additions and another for refactoring. Maybe it wasn't the best idea.

app/views/images/index.html.erb
bboe: Nevermind, I just realized there was one more commit.

app/controllers/images_controller.rb
bboe: In this case I think making the updates and the updates to the test make the same sense since it's not really a refactor.

Regarding, the next two commits, I like the way you did that because that does involve a refactor.

kkajla12: Alright, I'll add in those changes on this branch then

bboe: Should they be here now?

kkajla12: Oops, forgot to squash into that commit

kkajla12: @bboe , just added the update to the test into this commit

test/controllers/images_controller_test.rb
```diff
@@ -42,7 +42,7 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     post images_path('image' => {
       'url' => url
     })
-    assert_response :found
```
bboe: Was this to be replaced with an assert_redirected_to?

kkajla12: Yes, but that's technically part of a separate issue. Should I just address it in this branch?

kkajla12: Yes, I was supposed to keep that line here in this branch because technically that change is part of the refactor branch I opened for this issue. I shouldn't have deleted the line though. I'm not sure how that happened. I'll fix it

bboe:
👍

test/controllers/images_controller_test.rb
kkajla12: I had to force push to squash the change into this commit. Sorry for the inconvenience.

PR #28, opened 2016-09-02
Ropes image tags

### LINKED ISSUE #7: Image Tags
#### As a user I want to add tags to images.

__Story__:

Our users are loving the ability to add images to our service. However, now
that there are more than a handful of images, it is apparent that we need some
way to filter images by certain categories. In order to do that, we first need
to label images in some way. It is time to add the ability for our users to tag
images when they add them.

__Acceptance criteria__:
- [x] When I add a new image, there is a form field to add tags.
- [x] After I've saved an image with tags, I see those tags displayed with the
  associated image.
- [x] My application uses the acts-as-taggable-on gem to provide the tagging
  functionality.

__Reference material__:
- https://github.com/mbleigh/acts-as-taggable-on

__Dependencies__:
- Image Index

kkajla12:
Closes #7


kkajla12:
While working on this issue, I realized that my test files use different styling, and I would like to add a commit which refactors all of my test files to all use the rails 5 test style for consistency. Would it be ok to put that in this PR?

shfung:
I would recommend refactoring all of the test files in a separate PR.  The advantages of that is the new PR can be merged really quickly since it should only change the format of the tests and it doesn't change any of the functionality of your app.  Then all your existing PRs can rebase once that new PR is merged and any new test should be written in the new style.
Thought?

kkajla12:
I think that's a good idea. I'll go ahead and do that then, so please hold off on this PR for now. I'll message you on slack when the tests are refactored and this branch is rebased and ready to go.

db/migrate/20160829221233_acts_as_taggable_on_migration.acts_as_taggable_on_engine.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    create_taggings_table
+
+    add_index :taggings, :tag_id
+    add_index :taggings, [:taggable_id, :taggable_type, :context]
+  end
+
+  def self.create_taggings_table
+    create_table :taggings do |t|
+      t.references :tag
```
shfung: Maybe we can add some not-null constraints to some of the columns?

test/controllers/images_controller_test.rb
```diff
@@ -49,6 +68,19 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    url = 'http://images/1'
+    assert_difference 'Image.count', 1 do
+      post images_path(image: {
+        url: url,
+        tag_list: %w(0 1 2)
+      })
+    end
+    assert_redirected_to image_path(Image.last)
+    assert_equal 'Image created successfully!', flash.notice
+    assert_not_nil Image.find_by(url: url)
```
shfung: Maybe also assert on the content of the tags to make sure they are actually saved in the controller action?

test/controllers/images_controller_test.rb
```diff
@@ -49,6 +68,19 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_not_nil Image.find_by(url: url)
   end
 
+  test 'should create image with tags' do
+    url = 'http://images/1'
+    assert_difference 'Image.count', 1 do
+      post images_path(image: {
+        url: url,
+        tag_list: %w(0 1 2)
```
shfung: Since you have added `tag_list` to your form, I think we should include the key `tag_list` to all the post requests to the create action in all the tests.

test/flows/image_flow_test.rb
```diff
@@ -23,4 +23,29 @@ class ImageFlowTest < FlowTestCase
...
+
+    assert_equal 'Images', page.find('h1').text
+    page.click_link('Add Image')
+
+    page.find('#image_url').set('')
+    page.find('#image_tag_list').set('0, 1')
+    page.click_button('Create Image')
+    assert has_current_path? images_path
+    assert_equal 'Upload a new image URL!', page.find('h1').text
+    assert_equal "can't be blank", page.find('span').text
```
shfung: The above test snippet seems to be duplicated with the one in `default_site_flow` test.  Maybe we don't need to retest the error case that is not related to tagging?
Alternatively, you can add tests for these two options in your form: 
```
error_html: { class: 'text-danger' },
input_html: { value: @image.tag_list.to_s } 
```
but they should be in separate tests.

kkajla12: Would it make sense to just combine the two flow tests into one test that goes through the entire site flow and creates an image with a url AND tags?

shfung: I think it makes sense to combine the tests.

test/models/image_test.rb
```diff
@@ -20,4 +20,13 @@ class ImageTest < ActiveSupport::TestCase
     assert_predicate image, :invalid?
     assert_equal 'is not a valid URL', image.errors.messages[:url].first
   end
+
+  def test_url_and_tags__valid
```
shfung: Use new test naming style?

db/migrate/20160829221233_acts_as_taggable_on_migration.acts_as_taggable_on_engine.rb
kkajla12: Do references and indexes have to be explicitly specified as not-nullable or is that implicitly enforced by Rails?

test/models/image_test.rb
```diff
@@ -20,4 +20,13 @@ class ImageTest < ActiveSupport::TestCase
     assert_predicate image, :invalid?
     assert_equal 'is not a valid URL', image.errors.messages[:url].first
   end
+
+  test 'url_and_tags_valid' do
```
shfung: I found it a little weird that while you are using the new test naming style, you still put `_` in between words.  Is there any reason why you are keeping the `_`? If not, do you mind removing them?

db/migrate/20160829221233_acts_as_taggable_on_migration.acts_as_taggable_on_engine.rb
```diff
@@ -2,7 +2,7 @@
 class ActsAsTaggableOnMigration < ActiveRecord::Migration
   def self.up
     create_table :tags do |t|
-      t.string :name
+      t.string :name, null: false
```
shfung: NIce! One more comment about these changes is that all the changes of a single migration file should be contained within one commit, so that developer won't run into situations where a migrated migration file is actually stale and the latest changes to the file are not picked up. Can you combine this commit with https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/28/commits/41074ba83feaec8d63f9b0abf225c5eda2c7cb0f once we are done with code review?

test/models/image_test.rb
kkajla12: I think that's weird too. I'm not sure why I didn't change it. Is that something I can add to this PR?

db/migrate/20160829221233_acts_as_taggable_on_migration.acts_as_taggable_on_engine.rb
kkajla12: Yes, I will do that! I was thinking the same thing so I marked all of the commits that I want to combine with previous commits with the appropriate commit message telling me which commit to squash into.

shfung:
:+1:

PR #31, opened 2016-09-06
Ropes add style challenge exercises

kkajla12:
Extension of Issue #6


app/assets/stylesheets/content.scss
```diff
@@ -0,0 +1,10 @@
+.content {
+  display: flex;
+  height: 100vh;
+  min-height: 100vh;
+  flex-direction: column;
```
aaronpanch: Alphabetize :)

app/assets/stylesheets/content.scss
```diff
@@ -0,0 +1,10 @@
+.content {
+  display: flex;
+  height: 100vh;
```
aaronpanch: I don't think you need _both_ the height and min-height.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,9 @@
 module ApplicationHelper
+  def navbar_link(content, path)
+    li_class = current_page?(path) ? 'nav-item active' : 'nav-item'
```
aaronpanch: This is great -- I just have one minor comment, and believe me it's minor-- an ever so slightly less redundant way of writing this would be to only append the active class on to the base nav item, and that is easy when using an array.  So star with `li_classes = ['nav-item']` and append `active` when it's current.  As a bonus, `content_tag` will (magically ;) join the array with spaces.

app/assets/stylesheets/content.scss
kkajla12: ...so close. Missed it by that much. Resolved in 664e9bc57699da2a84d97cd4fa107ad897045fbb

app/assets/stylesheets/content.scss
kkajla12: Resolved in 664e9bc57699da2a84d97cd4fa107ad897045fbb

app/helpers/application_helper.rb
kkajla12: Resolved in 91a344d6f0de313c641c66ada693a09da1effcd5. Looks good?

app/assets/stylesheets/content.scss
```diff
@@ -0,0 +1,9 @@
+.content {
+  display: flex;
+  flex-direction: column;
+  height: 100vh;
```
aaronpanch: Ah I think `min-height` is the one you wanted.  I'm not 100% sure this works, but I wouldn't be surprised if it does.  We want to use min-height because then it allows the `content` element to grow when there is lots of content, or when there isn't much, it maintains a minimum height.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,9 @@
 module ApplicationHelper
+  def navbar_link(content, path)
+    li_class = ['nav-item']
+
+    content_tag(:li, class: current_page?(path) ? li_class << 'active' : li_class) do
```
aaronpanch: As a personal preference, I'd prefer to see the the `li_class` get updated on it's own line, then the content tag doesn't need to "know" about it.  That is:

``` ruby
li_class = ['nav-item']
li_class << 'active' if current_page?(path)

content_tag(stuff..., class: li_class) do ...
```

PR #32, opened 2016-09-08
Ropes delete image

### LINKED ISSUE #8: Delete Image
#### As a user I want to delete an image.

__Story__:

Our users are awesome at adding images to our web service. Some of our users
add dozens in a single session. Occasionally, however, some users add the wrong
image, and would like a way to remove those images. Please provide our users
with the ability to delete images. Note that at this point we trust our users
not to abuse this feature and delete more images than they should.

__Acceptance criteria__:
- [x] The user is prompted for confirmation prior to deletion.
- [x] Deleted images are removed from the homepage.

__Dependencies__:
- Image Index

kkajla12:
Closes #8


app/views/images/show.html.erb
```diff
@@ -20,3 +20,7 @@
     <% end %>
   </div>
 </div>
+<div class="row flex-center">
+  <%= button_to 'Delete Image', { action: 'destroy', id: @image.id },
+                method: :delete, remote: true, data: { confirm: 'Are you sure?' }, class: 'btn btn-danger' %>
```
D-D-M: I liked your 'confirm' message above a little bit more :) "Are you sure you want to delete this image?"

test/controllers/images_controller_test.rb
```diff
@@ -96,4 +96,12 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
     assert_select 'span', "can't be blank"
   end
+
+  test 'should delete image' do
+    image = Image.create!(url: 'http://image(s/0', tag_list: %w(0 1 2))
+    assert_difference 'Image.count', -1 do
+      delete image_path(image)
+    end
+    assert_response :no_content
+  end
```
D-D-M: This is a good test, but I think you should also probably have a flow test for your deletion flows. One to check deletion from the index page, and one to check deletion from the show page 👍

D-D-M: Also, when you write flow tests for the deletion, you'll probably have to [change your Capybara driver](https://github.com/AppFolioOnboarding/base/commit/95c16ab53c7064ebfc7fe7e244a974785da4c6c2#diff-e32affa1c612d33896f619dfcede19cfR30) from `Rack::Test` to a default Javascript driver, so that your "Are you sure?" confirmation modals will actually pop up (`Rack::Test` doesn't have Javascript turned on).

Once you have that set up, you can have your flow test...
1. Click the delete button
2. Click cancel on the modal
3. Verify that nothing got deleted
4. Click the delete button again
5. Click confirm on the modal
6. Verify that the correct image was deleted

💯

D-D-M:
LGTM 👍

app/views/images/index.html.erb
```diff
@@ -18,6 +18,14 @@
               <%= link_to(tag, images_path(tag_filter: tag), class: 'chip') %>
           <% end %>
         </div>
+        <div class="tile__buttons text-xs-center">
+          <%= button_to 'Delete',
```
rmacklin: I don't think there's any reason to use `button_to`...pretty much ever. The markup it generates is a ton of overkill:
```html
<form class="button_to" method="post" action="/images/20">
  <input type="hidden" name="_method" value="delete">
  <i class="btn btn-danger btn-sm js-delete-image-button waves-input-wrapper waves-effect waves-light" style="color:rgb(255, 255, 255);background:rgba(0, 0, 0, 0)">
    <input data-confirm="Are you sure you want to delete this image?" class="waves-button-input" type="submit" value="Delete" style="background-color:rgba(0,0,0,0);">
  </i>
  <input type="hidden" name="authenticity_token" value="SppGd5fnwvZYz02qDgkhSsAzGjdsUAKt/nX8ix8dM44kgFkD5bV+ljcNBmzEBGkgK2l10+eufMhrjt+c0QPGuA==">
</form>
```
`link_to` generates more sensible markup and should be sufficient:
```html
<a data-confirm="Are you sure you want to delete this image?" class="btn btn-danger js-delete-image-button" rel="nofollow" data-method="delete" href="/images/20">Delete</a>
```

rmacklin: cc @D-D-M

bboe: Hey @rmacklin I was the one that suggested button_to should be used. The reason is that links can be middle clicked or cmd/ctrl+clicked which results in unintended behavior for the users. Buttons do not have those actions. In this case it's not terrible that a middle click on the button opens a show page, but it's still undesired behavior.

In other cases, some actions which map to their own URL that do not have a GET action result in a tab opening which loads an error page (500 or 406). As a result, link_to with remote: true should probably only be used if the link actually also happens to work via a standard GET request.

rmacklin: As mentioned on Slack:
Command+clicking a delete button isn't something that makes sense conceptually. If users do that and get a show page, I don't think we care. For cases when it results in an exception, we can care, but this one doesn't.
I discovered this `button_to` while reviewing the Share Image PR where `button_to` was copied to an action that should 100% be using `link_to` (it's just a link to a new page, and command+click should be supported for it).
I'd say we should only avoid using `link_to` when it would generate exceptions if command+clicked.

kkajla12: @bboe how about ```link_to``` without ```remote: true```? We ended up removing the AJAX, so ```remote: true``` is no longer set on the buttons. In the case from the other PR that @rmacklin is referring to, I did incorrectly (in any case) use ```button_to``` where I should have used ```link_to``` since the purpose of the link is only to navigate to the share page, which is only a GET request.

bboe: I think `link_to` without `remote: true` is perfect. With `remote: true` is also fine so long as it also works when making a regular get request.

PR #35, opened 2016-09-08
Ropes images by tag

### LINKED ISSUE #9: Images by Tag
#### As a user I want to view all images associated with a tag.

__Story__:

Now that we have added the ability for images to be associated with tags, let
us provide the ability for users to filter images by a given tag. Nana, who
loves fast planes, will be delighted to see only images labeled with SR-71.

__Acceptance criteria__:
- [x] When I click on a tag, I see a filtered list of only the images that have
  the tag I clicked on.
- [x] The tag name, not its ID, is used to perform the filtering.

__Discussion topics__:
- [x] Where in the code should we implement this functionality?
- [x] What happens if a user tries to manually filter by a nonexistent tag?

__Dependencies__:
- Image Tags

kkajla12:
Closes #9


app/controllers/images_controller.rb
```diff
@@ -1,6 +1,11 @@
 class ImagesController < ApplicationController
   def index
-    @images = Image.order(created_at: :desc)
+    if params[:tag_filter]
+      @images = Image.tagged_with(params[:tag_filter]).order(created_at: :desc)
+      @images = @images.empty? ? Image.order(created_at: :desc) : @images
```
shiboying: If you cannot filter the images by a tag. It would be nice to tell the user about it(e.g. flash message) besides showing all images.

app/views/images/show.html.erb
```diff
@@ -13,8 +13,8 @@
       <div class="card-block">
         <div class="card-text">
           <% @image.tag_list.each do |tag| %>
-            <span class="chip"><%= tag %></span>
-          <% end %>
+            <%= link_to(tag, images_path + '?tag_filter=' + tag, class: 'chip') %>
```
shiboying: The easier way of getting the link could be ` images_path(tag_filter: tag)`

test/controllers/images_controller_test.rb
```diff
@@ -96,4 +96,38 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
   end
+
+  test 'should filter images by tag' do
+    Image.create!(url: 'http://images/0', tag_list: %w(A C))
+    Image.create!(url: 'http://images/1', tag_list: ['A'])
+    Image.create!(url: 'http://images/2', tag_list: %w(A C))
+    Image.create!(url: 'http://images/3', tag_list: %w(B C))
+    Image.create!(url: 'http://images/4', tag_list: ['B'])
+
+    get images_path(tag_filter: 'A')
```
shiboying: You could put an empty line between assertions and test the line of the code. It is easier to read.

test/controllers/images_controller_test.rb
```diff
@@ -96,4 +96,38 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+      assert_equal 3, images.size
+      images.each do |image|
+        assert_select image, '.chip' do |tags|
+          all_contain_tag = false
+          tags.each do |tag|
+            all_contain_tag = tag.text == 'A' || all_contain_tag
+          end
+          assert all_contain_tag
+        end
+      end
```
shiboying: I understand your assertion here. But I think it could be better to make it less complicated. What you want to assert is after you filter with 'A', there are only three pictures. You could just assert those three pictures links show. I think in the way it is easier to maintain the code.

app/controllers/images_controller.rb
kkajla12: Good idea! I should've added this from the beginning! Resolved in 5075a5c7e21895a8a83569badea958fda6a4af46

app/views/images/show.html.erb
kkajla12: Of course. I used the helper in the controller, so I'm not sure why I didn't think to use it here. Also, I think I should be linking to ```root_path(tag_filter: tag)``` instead because the root route goes to the images controller's index action. Right?

test/controllers/images_controller_test.rb
kkajla12: I see. That  makes sense, and it's much simpler. Resolved in a6df41451dbdaf165a68cd7ddca1d62aaab3f473

test/controllers/images_controller_test.rb
kkajla12: Resolved in a6df41451dbdaf165a68cd7ddca1d62aaab3f473

shiboying:
Look good to me.

PR #37, opened 2016-09-08
change order in which javascript loads and manually add mdb.js to loa…

kkajla12:


app/views/layouts/application.html.erb
```diff
@@ -20,7 +21,7 @@
       <%= render 'layouts/footer' %>
     </footer>
 
-    <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
+    <%= javascript_include_tag 'mdb', 'data-turbolinks-track': 'reload' %>
```
rmacklin: This isn't the correct approach. We don't want to load an extra external script because each one has a performance penalty. We don't need to use turbolinks, so if that was what was causing your problem, just remove turbolinks altogether instead.

PR #43, opened 2016-09-15
Flow tests with page objects

kkajla12:
Closes #38

PR #45, opened 2016-09-17
Ropes page objects fix

kkajla12:
The commits in this pull request fix a mistake in #43 whereby the flow test for deleting an image from the show page was implemented incorrectly.


app/views/images/show.html.erb
```diff
@@ -1,24 +1,27 @@
...
-          <% @image.tag_list.each do |tag| %>
-            <%= link_to(tag, images_path(tag_filter: tag), class: 'chip') %>
-        <% end %>
+        <div class="card-text">
+          <div class="row flex-center js-tags">
+            <% @image.tag_list.each do |tag| %>
+              <%= link_to(tag, images_path(tag_filter: tag), class: 'chip') %>
+            <% end %>
+          </div>
+          <br>
```
rmacklin: This `<br>` is a code smell. It's not semantic and it can most likely be replaced by adding an appropriate margin via CSS.

app/views/images/show.html.erb
```diff
@@ -1,24 +1,27 @@
...
+          <div class="row flex-center">
+            <%= button_to 'Delete',
+                          { action: 'destroy', id: @image.id },
+                          method: :delete,
+                          data: { confirm: 'Are you sure you want to delete this image?' },
+                          class: 'btn btn-danger' %>
+          </div>
```
rmacklin: It seems there is a fair amount of duplication between the show page and index page templates for displaying an image. It would be great if they could use a shared partial instead. That would make the fact that their page objects share an `ImageCard` element more natural and also more robust.

rmacklin: We can address this comment in a separate PR.

test/flows/images_crud_test.rb
```diff
@@ -40,6 +40,10 @@ class ImagesCrudTest < FlowTestCase
     image_to_delete = images_index_page.images.find do |image|
       image.url == ugly_cat_url
     end
+
+    image_show_page = image_to_delete.view!
+    image_to_delete = image_show_page.image
```
rmacklin: It would be a cleaner/simpler page object interface if we could avoid this variable, and just call `delete` and `delete_and_confirm!` directly on the `image_show_page` (like we were [in the page object skeleton](https://github.com/AppFolioOnboarding/image-sharer-kkajla12/blob/2d3c99608b3bbccb80124f71e2d86a310ceeb684/test/flows/images_crud_test.rb#L42-L49)). Even though you've moved the `delete` and `delete_and_confirm!` methods to `ImageCard` , this can still be accomplished with `delegate`:
```ruby
delegate :delete, :delete_and_confirm!, to: :image
```

test/flows/images_crud_test.rb
```diff
@@ -17,7 +17,7 @@ class ImagesCrudTest < FlowTestCase
     image_show_page = new_image_page.create_image!
     assert_equal 'Image created successfully!', image_show_page.flash_message(:success)
 
-    assert_equal image_url, image_show_page.image_url
+    assert_equal image_url, image_show_page.image.url
```
rmacklin: Similar to my last comment, I think the `ShowPage` should still provide the `image_url` method in its interface so that callers don't have to go through the `image` element to get it.

test/page_objects/images/show_page.rb
```diff
@@ -3,14 +3,10 @@ module Images
...
 
       collection :tag_list, locator: '.js-tags', item_locator: '.chip'
 
-      def image_url
-        image.node[:src]
-      end
-
       def tags
         tag_list.map(&:text)
       end
```
rmacklin: It seems that `tag_list` and `tags` are both duplicated from `ImageCard` and could instead by `delegate`d to `image`

app/views/images/show.html.erb
kkajla12: Resolved in ba26de61e25324e025272bbe4f741a6a7d67120b. I just removed it for now. I'll refactor the image card into a separate partial in another PR which will fix this issue as well.

test/flows/images_crud_test.rb
kkajla12: Resolved in 04c43840a562d81c855307bcad12e02a08b6e6c9

test/flows/images_crud_test.rb
kkajla12: This is cool. Resolved in 04c43840a562d81c855307bcad12e02a08b6e6c9

test/page_objects/images/show_page.rb
kkajla12: Resolved in 04c43840a562d81c855307bcad12e02a08b6e6c9

test/flows/images_crud_test.rb
rmacklin: Not exactly. In that commit, the method is called `url` which is a bit too ambiguous (`image_show_page.url` sounds like it could return the URL of the show page, not the URL of the image being shown). My comment here was that the method would still be called `image_url`.

test/flows/images_crud_test.rb
```diff
@@ -17,7 +17,7 @@ class ImagesCrudTest < FlowTestCase
     image_show_page = new_image_page.create_image!
     assert_equal 'Image created successfully!', image_show_page.flash_message(:success)
 
-    assert_equal image_url, image_show_page.image.url
+    assert_equal image_url, image_show_page.url
```
rmacklin: See https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/45#r79502465

app/views/images/show.html.erb
kkajla12: I've pulled the image card out into a partial in a local branch. I'll submit a pull request for that once I merge this in and rebase that branch.

test/flows/images_crud_test.rb
kkajla12: Resolved in c26192e04d9184b7f8c6ff5f914c3d072849b05e

test/flows/images_crud_test.rb
kkajla12: Oh, looking back at your first comment that makes perfect sense. Not sure why I did it this way. Resolved in c26192e04d9184b7f8c6ff5f914c3d072849b05e

PR #46, opened 2016-09-19
Ropes refactor image tile into partial

kkajla12:
Pull the image tile component out into a partial to be used in the index and show pages.


app/views/images/index.html.erb
```diff
@@ -2,21 +2,7 @@
...
-        </div>
-        <div class="tile__buttons text-xs-center">
-          <%= button_to 'Delete',
-                        { action: 'destroy', id: image.id },
-                        method: :delete,
-                        data: { confirm: 'Are you sure you want to delete this image?' },
-                        class: 'btn btn-danger js-delete-image-button' %>
-        </div>
-      </div>
+      <%= render partial: 'layouts/image_tile', locals: { image: image } %>
```
rmacklin: Why did you choose to put the partial in layouts?

rmacklin: On a related note, if you use some "conventional" naming for partials, Rails gives you a nice shorthand. If this partial was named `_image`, then rather than specifying the partial you could just call `render image`:
```erb
<% @images.each do |image| %>
  <%= render image %>
<% end %>
```
and since you're doing that for the whole `@images` collection, it can be reduced to:
```erb
<%= render @images %>
```

app/views/images/show.html.erb
```diff
@@ -1,26 +1,4 @@
...
-                          method: :delete,
-                          data: { confirm: 'Are you sure you want to delete this image?' },
-                          class: 'btn btn-danger' %>
-          </div>
-        </div>
-      </div>
-    <% end %>
-  </div>
+<div class="flex-center">
+  <%= render partial: 'layouts/image_tile', locals: { image: @image } %>
```
rmacklin: Following the previous comment, this could just be `<%= render @image %>`

app/views/images/index.html.erb
kkajla12: I guess I thought it might be useful across the application since it's the focus of the app, but then again, it is an image. In which case, it just belongs as a partial within the scope of the images controller and views.

kkajla12: Wow, that reduces the markup to almost nothing compared to before.

app/views/images/show.html.erb
kkajla12: Resolved in 55a5e0432ae2a6f8bbdbf12b1c1dec4bee9739c1

rmacklin:
The new commit looks good, but this needs a rebase.

kkajla12:
Yup! I'll rebase after squashing and before merging!

PR #48, opened 2016-09-20
Ropes share image

### LINKED ISSUE #10: Share Image
#### As a user I want to share an image with someone via email.

__Story__:

This application is called Image Sharer. Sure, you can share images by adding
them to our web service, but how will people who do not know about this web
service be able to see those images? In phone calls with our customers we have
discovered that many of our users are manually emailing their friends links to
the images. That is awesome! We can make things more efficent for our users by
directly supporting sharing images through email.

__Acceptance criteria__:
- [x] There is a link to share an image.
- [x] After clicking on this link, a bootstrap modal appears with a form to
  send the email.
- [x] The form has fields for the recipient's email and an optional message.
- [x] The email we send has an HTML part and a text part.
- [x] The email contains the image, the custom message if provided, and a link
  inviting the recipient to check out our application.
- [x] The modal closes and a success message is shown if the email was sent
  successfully.
- [x] The user is alerted of errors if the operation was unsuccessful.

__Note__: This story is large, so it makes sense to break the implementation
down into more manageable pieces (a la elephant carpaccio). Please start by
implementing a basic sharing feature that uses a new page (not a modal), and
open a PR for that. Then, once we've verified the simplified version is
working, refactor it to use a modal, and open a second PR for that. The
[video](https://sites.google.com/a/appfolio.com/eng/new-hire-resources/engineering-academy/ropes/image-sharing-project/basics-of-bootstrap-modals-js-and-ajax)
provides some background and code examples that will be really helpful for the
second PR, so it's worth watching (and following along) before you begin that
part.

__Discussion topics__:
- [x] How could we develop the front end and back end separately? (And why
  might that be a good idea?)
- [x] How should we implement a non-CRUD action?
- [x] What are the different types of errors we should handle?
- [x] What is AJAX? What libraries can help us add AJAX to our Rails app?
- [x] How should we go about sending emails from our Rails app?
- [x] How could we show a preview of the image being shared in the modal?
- [x] What are the implications of allowing users to send emails from our app?

__Reference material__:
- http://guides.rubyonrails.org/action_mailer_basics.html
- https://elements.heroku.com/addons/sendgrid
- https://devcenter.heroku.com/articles/sendgrid#ruby-rails
- https://github.com/rails/jquery-ujs/wiki
- http://v4-alpha.getbootstrap.com/components/modal
- https://sites.google.com/a/appfolio.com/eng/new-hire-resources/engineering-academy/ropes/image-sharing-project/basics-of-bootstrap-modals-js-and-ajax

__Dependencies__:
- Image Index

kkajla12:
First part of #10 (no modal)


rmacklin:
In this case you don't want to say "Closes #10" because this is only the first part, so you don't want the issue to be automatically closed when this PR is merged. You can edit the description to say "First part of #10 (no modal)" and it will still link to the issue without automatically closing it upon merge.

app/controllers/images_controller.rb
```diff
@@ -41,9 +43,39 @@ def destroy
     render plain: 'Oops! The image you are attempting to delete was not found!', status: :not_found
   end
 
+  def share_form
+    @image = Image.find(params[:id])
```
rmacklin: @HaominZhang will probably comment that you could extract this into a `before_action`. And you could, but I don't particularly care too much if you do right now, since the finder ( `Image.find(params[:id])` ) is so simple.

app/views/images/_image.html.erb
```diff
@@ -5,11 +5,19 @@
...
   </div>
-  <div class="tile__buttons text-xs-center">
-    <%= button_to 'Delete',
-                  { action: 'destroy', id: image.id },
-                  method: :delete,
-                  data: { confirm: 'Are you sure you want to delete this image?' },
-                  class: 'btn btn-danger js-delete-image-button' %>
+  <div class="tile__buttons row">
+    <div class="col-md-6 text-xs-right">
+      <%= button_to 'Delete',
```
rmacklin: I see that this was using `button_to` since PR #32, so I left this comment there: https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/32#discussion_r79744956 (and CC'ed David since he reviewed that one).

app/views/images/_image.html.erb
```diff
@@ -5,11 +5,19 @@
...
+                    { action: 'destroy', id: image.id },
+                    method: :delete,
+                    data: { confirm: 'Are you sure you want to delete this image?' },
+                    class: 'btn btn-danger btn-sm js-delete-image-button' %>
+    </div>
+    <div class="col-md-6 text-xs-left">
+      <%= button_to 'Share',
```
rmacklin: We should avoid using `button_to` here as well. 
```erb
<%= link_to 'Share', share_form_image_path(image), class: 'btn btn-primary btn-sm' %>
```
should be sufficient.

app/views/images/share_form.html.erb
```diff
@@ -0,0 +1,18 @@
+<h1 class="flex-center">Share This Image</h1>
+<div class="row flex-center">
+  <div class="col-md-6">
+    <%= simple_form_for @form, url: share_image_path(@image), html: { novalidate: true } do |f| %>
+        <%= f.input :recipient,
+                    label: 'Send To',
+                    required: false,
```
rmacklin: Why are you setting this required attribute as not required?

app/views/images_mailer/share_image_email.html.erb
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <div class="image">
+      <%= image_tag(@image.url, class: 'image-fluid') %>
+    </div>
+    <div class="message">
```
rmacklin: Can we prepend the message with something like "Your friend said:" so it's clear that our application isn't responsible for the contents of that part of the email?

And if the `message` is optional, do we still want to display this paragraph? Is there a better way to handle the empty message case and not include this extra paragraph?

app/views/layouts/_confirmation_modal.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="modal fade" id="confirmationModal" tabindex="-1" role="dialog" aria-labelledby="confirmationModalLabel"
```
rmacklin: What is this file for? I don't see it used anywhere.

test/controllers/images_controller_test.rb
```diff
@@ -138,4 +138,56 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    </div>
+    <div class=\"message\">
+      <p>Hello, World!</p>
+    </div>
+    <br>
+    <p>If you haven't already, try out <a href=\"http://www.example.com/\">Image Sharer</a> for yourself!</p>
+    <p>From everyone at Image Sharer, have a great day!</p>
+  </body>
+</html>
+"
```
rmacklin: It's only necessary to test the full email content in one place: the mailer test. In the controller test we can just assert that the email that was delivered included the essential pieces. Otherwise changes to the mailer markup would require changing the mailer test _and the controller test_, even though the change has nothing to do with the controller logic.

test/controllers/images_controller_test.rb
```diff
@@ -138,4 +138,56 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_redirected_to images_path
     assert_equal 'Image deleted successfully!', flash[:success]
   end
+
+  test 'should get share image form' do
+    image = Image.create!(url: 'http://images/0', tag_list: %w(0 1 2))
+    get share_form_image_path(image)
+    assert_response :ok
```
rmacklin: We should also make assertions about the content of the page (e.g. that the share form was rendered).

test/controllers/images_controller_test.rb
```diff
@@ -138,4 +138,56 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    end
+
+    share_email = ActionMailer::Base.deliveries.last
+
+    assert_equal 'donotreply@appfolio.com', share_email.from[0]
+    assert_equal 'test@test.com', share_email.to[0]
+    assert_equal 'Someone Shared an Image With You Using Image Sharer!', share_email.subject
+    assert_equal text_body, share_email.text_part.body.to_s
+    assert_equal html_body, share_email.html_part.body.to_s
+  end
```
rmacklin: There are missing controller test cases. What other code paths have not been covered?

test/flows/images_crud_test.rb
```diff
@@ -107,4 +107,34 @@ class ImagesCrudTest < FlowTestCase
...
+    ugly_cat_url = 'http://www.ugly-cat.com/ugly-cats/uglycat041.jpg'
+    Image.create!([
+      { url: cute_puppy_url, tag_list: 'puppy, cute' },
+      { url: ugly_cat_url, tag_list: 'cat, ugly' }
+    ])
+
+    images_index_page = PageObjects::Images::IndexPage.visit
+    assert_equal 2, images_index_page.images.count
+    assert images_index_page.showing_image?(url: ugly_cat_url)
+    assert images_index_page.showing_image?(url: cute_puppy_url)
```
rmacklin: Are these two assertions necessary for this test?

test/flows/images_crud_test.rb
```diff
@@ -107,4 +107,34 @@ class ImagesCrudTest < FlowTestCase
...
+
+    image_to_share = images_index_page.images.find do |image|
+      image.url == cute_puppy_url
+    end
+
+    share_image_page = image_to_share.share
+    message = 'Hello, World!'
+    share_image_page = share_image_page
+      .share_image!(recipient: 'invalid', message: message).as_a(PageObjects::Images::SharePage)
+    assert_equal 'is invalid', share_image_page.recipient.error_message
```
rmacklin: You should also assert that the previous input values were preserved when the form was rerendered.

test/flows/images_crud_test.rb
```diff
@@ -107,4 +107,34 @@ class ImagesCrudTest < FlowTestCase
...
+
+    recipient = 'test@test.com'
+    share_image_page.recipient.set(recipient)
+
+    images_index_page = share_image_page.share_image!
+    assert_equal 'Image shared successfully!', images_index_page.flash_message(:success)
+  end
```
rmacklin: What other case should be tested?

test/mailers/images_mailer_test.rb
```diff
@@ -0,0 +1,45 @@
...
+    text_body = "http://images/0
+=========================================================
+
+Hello, World!
+
+If you haven't already, try out http://www.example.com/ for yourself!
+
+From everyone at Image Sharer, have a great day!
+"
+    html_body = "<!DOCTYPE html>
```
rmacklin: It's idiomatic ruby to use heredocs for this kind of thing.

app/models/share_image_form.rb
```diff
@@ -0,0 +1,5 @@
+class ShareImageForm
+  include ActiveModel::Model
+  attr_accessor :recipient, :message
+  validates :recipient, email: true
```
rmacklin: Consider declaring an explicit `presence` validation, even if `email: true` implicitly includes one (I'm not sure if it does).

rmacklin:
I've added some initial comments. @HaominZhang is also going to become a reviewer for this story, so he may add some more comments.

app/views/images/share_form.html.erb
```diff
@@ -0,0 +1,18 @@
+<h1 class="flex-center">Share This Image</h1>
+<div class="row flex-center">
+  <div class="col-md-6">
```
rmacklin: When I am on this page, how do I know what image I am sharing?

app/views/images_mailer/share_image_email.html.erb
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
```
rmacklin: This HTML wrapping should come from a shared mailer layout, and omitted from individual mailer templates

app/views/images_mailer/share_image_email.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <div class="image">
```
rmacklin: Your CSS files are not shipped with emails (by design), so this `class` won't do anything in your email (same for the other classes you're setting). You'll need to use inline styles directly on the elements or define classes inside a `<style>` tag which should be inserted in the `<head>`

app/controllers/images_controller.rb
```diff
@@ -41,9 +43,39 @@ def destroy
...
+    @form = ShareImageForm.new
+
+  rescue ActiveRecord::RecordNotFound
```
rmacklin: `rescue` should be reserved for _exceptional_ behavior rather than for normal control flow. We expect that users will run into this code path, so we should handle it with regular conditional logic (`if/else`) rather than exception handling (`rescue`). You can use `Image.find_by(id: params[:id])` instead which will return `nil` rather than raising an exception when the image is not found.

test/flows/images_crud_test.rb
```diff
@@ -107,4 +107,34 @@ class ImagesCrudTest < FlowTestCase
     images_index_page = images_index_page.clear_tag_filter!
     assert_equal 3, images_index_page.images.count
   end
+
+  test 'share an image' do
```
rmacklin: Sharing is not part of the CRUD operations, so we shouldn't add this test in this file. We can create a separate flow test file which will hold all our sharing flow tests

test/page_objects/images/image_card.rb
```diff
@@ -25,6 +25,11 @@ def delete_and_confirm!
         delete(&:accept)
         window.change_to(IndexPage)
       end
+
+      def share
```
rmacklin: By our conventions, this method should end in a `!` since it returns a new page object.

test/models/share_image_form_test.rb
```diff
@@ -0,0 +1,21 @@
...
+  test 'recipient is valid email' do
+    form = ShareImageForm.new(recipient: 'test@test.com', message: 'Hello, World!')
+
+    assert_predicate form, :valid?
+  end
+
+  test 'recipient is invalid email' do
+    form = ShareImageForm.new(recipient: 'invalid', message: 'Hello, World!')
+
+    assert_predicate form, :invalid?
```
rmacklin: You should assert the form errors when for invalid cases.

test/models/share_image_form_test.rb
```diff
@@ -0,0 +1,21 @@
+require 'test_helper'
+
+class ShareImageFormTest < ActiveSupport::TestCase
+  test 'recipient is valid email' do
+    form = ShareImageForm.new(recipient: 'test@test.com', message: 'Hello, World!')
+
+    assert_predicate form, :valid?
+  end
+
+  test 'recipient is invalid email' do
```
rmacklin: What about testing the empty recipient case?

test/flows/images_crud_test.rb
```diff
@@ -107,4 +107,34 @@ class ImagesCrudTest < FlowTestCase
...
+    cute_puppy_url = 'http://ghk.h-cdn.co/assets/16/09/980x490/landscape-1457107485-gettyimages-512366437.jpg'
+    ugly_cat_url = 'http://www.ugly-cat.com/ugly-cats/uglycat041.jpg'
+    Image.create!([
+      { url: cute_puppy_url, tag_list: 'puppy, cute' },
+      { url: ugly_cat_url, tag_list: 'cat, ugly' }
+    ])
+
+    images_index_page = PageObjects::Images::IndexPage.visit
```
rmacklin: We're currently testing sharing from the index page. We could also add a flow test for sharing from the show page, but I don't think it's necessary at this point: We could have a lighter test in our controller test that just verifies that the share button is rendered on the show page and has the correct link.

app/views/images_mailer/share_image_email.html.erb
```diff
@@ -0,0 +1,17 @@
...
+      <%= image_tag(@image.url, class: 'image-fluid') %>
```
rmacklin: Also...we should probably have some opening text in the email (both HTML and text parts) before the image, for context beyond the subject line (which can easily be skipped over).

app/views/images_mailer/share_image_email.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    </div>
+    <div class="message">
+      <p><%= @message %></p>
+    </div>
+    <br>
+    <p>If you haven't already, try out <%= link_to 'Image Sharer', root_url %> for yourself!</p>
+    <p>From everyone at Image Sharer, have a great day!</p>
```
rmacklin: Thanks! 😄

rmacklin:
Sending emails from the review app is currently broken. I'm guessing that's why you opened #49  (possibly after discussing with Ryan and him recalling me mentioning that sendgrid configuration should be its own PR, which it should). That's great, but we'll have to rebase this on top of that if we want to QA the email deliver.

app/controllers/images_controller.rb
HaominZhang: I agree with @rmacklin that since this is pretty simple, we don't need to extract it into a `before_action` but we should also keep in mind in more complex cases we need to do it

Gemfile
```diff
@@ -73,3 +73,6 @@ gem 'acts-as-taggable-on', '~> 4.0'
```
kkajla12: This comment has been addressed by #51. The corresponding commit (eeb43cf500c8a8e8dae44fbab47541cdaf9084cf) on this branch will be removed before the branch is merged into master.

app/views/images/share_form.html.erb
kkajla12: @rmacklin, When a form field is set as required, Simple Form adds an asterisk with a tooltip which doesn't work well with the material design front end framework I'm using. Here, the field is still required since the validations still run, so it's not possible to submit the form without this field populated.

app/views/images_mailer/share_image_email.html.erb
kkajla12: I added the classes purely to distinguish between the parts of the email thinking that I would use the selectors to select each part individually. However, since I just checked the entire content of the email in my tests, they aren't necessary in any case. I did mistakenly put the ```image-fluid``` class on the image though.

app/views/layouts/_confirmation_modal.html.erb
kkajla12: Oops, this is a modal I was using for extending the delete image story with a modal. Need to be careful with ```git add -A```

app/views/images_mailer/share_image_email.html.erb
rmacklin: Thanks for the explanation. Using the `id` attribute would be much more appropriate for that use case.

app/views/images/share_form.html.erb
rmacklin: So will you have to add `required: false` to every required form field? That's not ideal, is it? Simple form is highly configurable, though, and I think you can change the appearance of required fields, including the asterisk. This would be better for two reasons: 1) you don't have to set `required: false` for all required fields just because the style produced doesn't fit your app, and 2) having _some_ style differentiation between required and optional form fields is a very important usability enhancement that apps should try to keep at all costs.

app/views/images_mailer/share_image_email.html.erb
kkajla12: That would make more sense. I could also have used ```js-class-name``` since I used classes, right?

rmacklin: You could, but these sections of the email would be unique so an `id` is still a better fit. Generally, prefer `id`s unless the thing you're adding a locator/selector for isn't unique (e.g. items in a collection).

app/models/share_image_form.rb
kkajla12: The email_validator gem does perform a presence validation as well, but I liked the idea of adding one explicitly so that I could have it generate a different error message, since 'is invalid' doesn't really tell you that the field is empty. Resolved in 92955023b339e5701ea467e9c7334196a8553d36

app/models/share_image_form.rb
```diff
@@ -0,0 +1,5 @@
...
+end
```
kkajla12: Resolved in 92955023b339e5701ea467e9c7334196a8553d36. I also created a directory for form tests and moved the corresponding test file into that directory.

app/mailers/images_mailer.rb
```diff
@@ -0,0 +1,7 @@
+class ImagesMailer < ActionMailer::Base
```
kkajla12: I didn't use the generator, but I did follow [this](http://guides.rubyonrails.org/action_mailer_basics.html#sending-emails) tutorial. I just read through it again and it initially tells you to inherit from ```ActionMailer::Base``` if you don't want to use the generator, and then the rest of the code examples inherit from ```ApplicationMailer```

kkajla12: Resolved in b4325ee72e9571c963ae14add4a18e82ca2dc4c2

app/mailers/images_mailer.rb
```diff
@@ -0,0 +1,7 @@
...
+  def share_image_email(image, recipient, message)
```
kkajla12: Resolved in b4325ee72e9571c963ae14add4a18e82ca2dc4c2. I update the use of that function in later commits, but they will be squashed into the same commit, so I think it should be fine.

config/environments/development.rb
```diff
@@ -26,11 +26,15 @@
...
-  # Don't care if the mailer can't send.
-  config.action_mailer.raise_delivery_errors = false
-
   config.action_mailer.perform_caching = false
 
+  # Tell Action Mailer not to deliver emails to the real world.
+  config.action_mailer.delivery_method = :sendmail
+  config.action_mailer.perform_deliveries = true
+  config.action_mailer.raise_delivery_errors = true
+  config.action_mailer.default_options = { from: 'donotreply@appfolio.com' }
```
kkajla12: Resolved in b4325ee72e9571c963ae14add4a18e82ca2dc4c2

config/environments/development.rb
```diff
@@ -26,11 +26,15 @@
...
+  config.action_mailer.default_url_options = { host: 'http://www.example.com' }
```
kkajla12: Resolved in b4325ee72e9571c963ae14add4a18e82ca2dc4c2

config/environments/test.rb
```diff
@@ -33,6 +33,10 @@
   # The :test delivery method accumulates sent emails in the
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
+  config.action_mailer.perform_deliveries = true
+  config.action_mailer.raise_delivery_errors = true
+  config.action_mailer.default_options = { from: 'donotreply@appfolio.com' }
```
kkajla12: Resolved in b4325ee72e9571c963ae14add4a18e82ca2dc4c2

config/environments/test.rb
```diff
@@ -33,6 +33,10 @@
...
+  config.action_mailer.default_url_options = { host: 'http://www.example.com' }
```
kkajla12: Resolved in b4325ee72e9571c963ae14add4a18e82ca2dc4c2

app/controllers/images_controller.rb
kkajla12: I ended up using ```before_action``` because it helped me shorten my controller action methods when rubocop started complaining about their length. Resolved in a181f620ec0b3771d5fe5fbcd596f4214e2b428e

app/controllers/images_controller.rb
```diff
@@ -41,9 +43,39 @@ def destroy
...
   def image_params
     params.require(:image).permit(:url, :tag_list)
   end
+
+  def image_share_form_params
+    params.require(:share_image_form).permit(:recipient, :message)
+  end
+
+  def set_mailer_url_options
+    ActionMailer::Base.default_url_options = { host: request.host_with_port }
```
kkajla12: I think you might be right. My tests have been fine since removing it. Resolved in ca82d739251fdfc67cffa12f6488c7443e8dc684

app/views/images/_image.html.erb
kkajla12: Resolved in 1a287abed677022ea7dba225b96be1e58da91c84

app/views/images/_image.html.erb
kkajla12: Since there isn't any exceptional behavior when the user cmd+clicks or middle-clicks on the delete button here, I changed it to a ```link_to```. Resolved in 0730f4a6ea95e112d945d33c05b2beb825bf7e96

app/views/images/share_form.html.erb
kkajla12: I added an image tile to display above the share form so that the user knows which image they are currently sharing. Resolved in 13a2153479226e070cd9d86fd5217f844cfa93ba

app/views/images_mailer/share_image_email.html.erb
kkajla12: I added a short sentence in the beginning to give context to the email and used a presence check on the message to only display that paragraph when a message is present. Resolved in 1207be45ee842a50034fb256c95c480c82e22075 and 49092cc80b087176bfba4eef01a8db3236202096

app/views/layouts/_confirmation_modal.html.erb
kkajla12: Deleted this file in ddc61616cb1dde902099a548d1713be9aeecf6d2

test/controllers/images_controller_test.rb
kkajla12: That makes a lot of sense. I should have done that from the beginning. Resolved in 1ee2a3826f9a2c69c43904779d873007956f686e

test/controllers/images_controller_test.rb
kkajla12: I added tests for attempting to share an non-existent image and having an empty recipient. Resolved in 1ee2a3826f9a2c69c43904779d873007956f686e

test/flows/images_crud_test.rb
kkajla12: I suppose not since it repeats what another test is already checking and also isn't the focus of this test. Resolved in 0da8e126ec70726dad49715b893576debf14da21

test/flows/images_crud_test.rb
kkajla12: Resolved in 0da8e126ec70726dad49715b893576debf14da21

test/flows/images_crud_test.rb
kkajla12: I added another flow test to test attempting to share a non-existent image. Resolved in a181f620ec0b3771d5fe5fbcd596f4214e2b428e

test/mailers/images_mailer_test.rb
kkajla12: Woah, heredocs are cool. Resolved in a181f620ec0b3771d5fe5fbcd596f4214e2b428e

app/views/images_mailer/share_image_email.html.erb
kkajla12: Resolved in 1207be45ee842a50034fb256c95c480c82e22075

app/views/images_mailer/share_image_email.html.erb
kkajla12: I added a short sentence in the beginning to give context to the email. Resolved in 1207be4

app/controllers/images_controller.rb
kkajla12: I added the use of ```Image.find_by(id: params[:id])``` in ca82d739251fdfc67cffa12f6488c7443e8dc684 and later added the use of a ```before_action``` in a181f620ec0b3771d5fe5fbcd596f4214e2b428e

test/flows/images_crud_test.rb
kkajla12: Resolved in 0da8e126ec70726dad49715b893576debf14da21

test/flows/images_crud_test.rb
kkajla12: I added a controller test to check if the share image button shows correctly on the show page with the correct url in 0da8e126ec70726dad49715b893576debf14da21 and 9e6d0510598f3805dd04397d66151d5025553c1e

test/page_objects/images/image_card.rb
kkajla12: Resolved in a181f620ec0b3771d5fe5fbcd596f4214e2b428e

test/models/share_image_form_test.rb
kkajla12: Resolved in 49660f4aa1063111076d2ea3bbc98cbf85ffe2b1

test/models/share_image_form_test.rb
kkajla12: Resolved in 49660f4aa1063111076d2ea3bbc98cbf85ffe2b1

app/views/images/share_form.html.erb
kkajla12: @rmacklin, it is not ideal at all. Can I work on updating the simple form default configuration and updating my forms throughout the app in a separate PR? I think it's an independent change.

HaominZhang: I think if you want to avoid the asterisk, you can simply remove the `required:` attribute instead of setting its value to be `false`. @rmacklin Do you think this is preferable?

kkajla12: I actually went back and removed the required attribute and the result is perfect. I think I was having some issues with my frontend framework's JavaScript loading initially. I was since able to fix that issue and a side-effect of that was fixing this. I will remove the assignment of required to false. Thanks for point that out @HaominZhang

HaominZhang: Although this could be a fix, I am wondering why you want to avoid the asterisk for this required field. What I am thinking is we should make the front end consistent with backend so if the field is required in backend, from user experience, it would be better to show the asterisk to let user know this is a required field. What's your concern about it?

kkajla12: @HaominZhang, sorry, my previous comment was unclear. Your first comment here reminded me that I don't need to specify ```required: false``` anymore and can just allow the default (which is ```required: true```) to apply. Like I said in my previous comment, my frontend framework was initially clashing with the way Simple Form adds the asterisk, but since I fixed that problem, I can now allow Simple Form to apply the asterisk and it looks good now. I hope that clears things up.

HaominZhang: Oh I got it! Thanks for the clarification!

kkajla12: Resolved in 4e0beaba3424f1249481b14ef678b9accdb94615

app/views/images_mailer/share_image_email.html.erb
```diff
@@ -0,0 +1,13 @@
+<div>
```
rmacklin: @aaronpanch may have already discussed this with you, but `<div>`s with no attributes are a code smell and should be avoided.

test/controllers/images_controller_test.rb
```diff
@@ -138,4 +138,92 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+
+    assert_no_difference 'ActionMailer::Base.deliveries.size' do
+      post share_image_path(image, share_image_form: {
+        recipient: '',
+        message: 'Hello, World!'
+      })
+    end
+
+    assert_response :unprocessable_entity
+    assert_select 'span', "can't be blank"
```
rmacklin: Consider a more specific selector

test/controllers/images_controller_test.rb
```diff
@@ -138,4 +138,92 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    assert_redirected_to images_path
+    assert_equal 'Oops! The image you are attempting to share was not found!', flash[:error]
+  end
+
+  test 'should show share button on show page' do
+    image = Image.create!(url: 'http://images/0', tag_list: %w(0 1 2))
+
+    get image_path(image)
+    assert_response :ok
+    assert_select '.js-image-share-button', href: share_image_path(image)
```
rmacklin: If you change `href` to something else, does this test fail?

test/controllers/images_controller_test.rb
```diff
@@ -138,4 +138,92 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+        message: 'Hello, World!'
+      })
+    end
+
+    share_email = ActionMailer::Base.deliveries.last
+
+    assert_equal 'donotreply@appfolio.com', share_email.from[0]
+    assert_equal 'test@test.com', share_email.to[0]
+    assert_equal 'Someone Shared an Image With You Using Image Sharer!', share_email.subject
+
```
rmacklin: Perhaps what I said in https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/48#discussion_r79745742 was misunderstood:
>  In the controller test we can just assert that the email that was delivered included the essential pieces.

That is, you still want to make a few basic assertions of the important parts of the email body.

test/flows/images_share_test.rb
```diff
@@ -0,0 +1,51 @@
...
+    assert_equal message, share_image_page.message.text
+
+    recipient = 'test@test.com'
+    share_image_page.recipient.set(recipient)
+
+    images_index_page = share_image_page.share_image!
+    assert_equal 'Image shared successfully!', images_index_page.flash_message(:success)
+  end
+
+  test 'try to share an invalid image' do
```
rmacklin: This is a nonexistent image rather than an invalid image.

app/mailers/images_mailer.rb
rmacklin: FWIW, I think it's a good experience to give the generator a try, even if you don't end up using it on this branch.

kkajla12: Alright, I'll take your tip and start using it from now on. It'll save me from creating the file and typing the code to setup the class every time. I use it for generating migrations, but I haven't used it for things like controllers or models because I feel like I've been primed to think that it creates extra files that bloat the project even though it doesn't in such a use case.

app/views/images_mailer/share_image_email.html.erb
kkajla12: My only intention for using divs here was to get a new line between those two sections of the email, but I ignored the fact that the p tags will get their own lines anyways, so I don't need the divs.

kkajla12: Resolved in e733238623aa094d0abfbe8aa5ddaf59886b5dfe

test/controllers/images_controller_test.rb
kkajla12: Yeah, I did misunderstand that. I added in checks of the image url, the email message, and the link to the application. Resolved in aeb430a11c11c6c3fb3267d8b801ec5f156a7271

test/controllers/images_controller_test.rb
kkajla12: It didn't fail. I guess I just took that link for granted. What exactly is it doing then?

kkajla12: Resolved in af401db8dbc9c1702d6b714e2a9986282dd1c302

test/controllers/images_controller_test.rb
kkajla12: Resolved in 5fa6793fe73c4cbb49339f36718507ab5c8c5da9 and d3589441d6484724a8433d218b0d2ebf2d4eed46

test/flows/images_share_test.rb
kkajla12: Resolved in b9d36a3730146127a4de4b5706881114f3011f36

test/controllers/images_controller_test.rb
```diff
@@ -174,6 +175,11 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_equal 'donotreply@appfolio.com', share_email.from[0]
     assert_equal 'test@test.com', share_email.to[0]
     assert_equal 'Someone Shared an Image With You Using Image Sharer!', share_email.subject
+    assert_select_email do
+      assert_select 'img[src=?]', 'http://images/0'
+      assert_select '#email-message', message
+      assert_select '#email-link[href=?]', 'http://localhost:3000/'
+    end
```
rmacklin: Neat, I didn't know about `assert_select_email`. However, I suspect this only tests the HTML part of the email, but we should also check that the text part contains these essential parts.

test/controllers/images_controller_test.rb
rmacklin: Yeah, it's just ignored. Ruby methods that take an `options` hash typically just ignore any unknown keys. This is an advantage of using keyword arguments over an options hash: if you pass an invalid keyword argument an error will be raised.

rmacklin: A good takeaway here though is to make sure you know your assertion is testing what you think it's testing.

app/views/images/_image.html.erb
```diff
@@ -5,11 +5,16 @@
...
-  <div class="tile__buttons text-xs-center">
-    <%= button_to 'Delete',
-                  { action: 'destroy', id: image.id },
-                  method: :delete,
-                  data: { confirm: 'Are you sure you want to delete this image?' },
-                  class: 'btn btn-danger js-delete-image-button' %>
+  <div class="tile__buttons row">
+    <div class="col-md-6 text-xs-right">
+      <%= link_to 'Delete',
+                    { action: 'destroy', id: image.id },
```
rmacklin: This can just be `image_path`

test/controllers/images_controller_test.rb
```diff
@@ -138,4 +138,98 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    assert_response :ok
+
+    assert_select 'h1', 'Share This Image'
+    assert_select '.new_share_image_form'
+  end
+
+  test 'should not show share form for invalid id' do
+    Image.create!(url: 'http://images/0', tag_list: %w(0 1 2))
+
+    get share_form_image_path(id: 2)
```
rmacklin: The convention is to use `-1` for the id of a nonexistent record

app/views/images/share_form.html.erb
```diff
@@ -0,0 +1,27 @@
...
+      <% end %>
+    </div>
+  </div>
+</div>
+<div class="row flex-center">
+  <div class="col-md-6">
+    <%= simple_form_for @form, url: share_image_path(@image), html: { novalidate: true } do |f| %>
+        <%= f.input :recipient,
+                    label: 'Send To',
+                    error_html: { class: 'text-danger', id: 'recipient-error' } %>
```
rmacklin: Our convention is that `id`s should use underscores (not hyphens) as the word delimiter

app/views/images_mailer/share_image_email.html.erb
```diff
@@ -0,0 +1,9 @@
+<p>Cool! Your friend shared this image with you:</p>
+<%= image_tag(@image.url) %>
+<% if @message.present? %>
+  <p>Your friend said:</p>
+  <p id="email-message"><%= @message %></p>
```
rmacklin: Our convention is that `id`s should use underscores (not hyphens) as the word delimiter

test/controllers/images_controller_test.rb
kkajla12: You're right, it doesn't check the plain-text part. I added some extra assertions to check that the plain text email contains the important strings we're expecting. Resolved in 7c8e29a5cce05f126e9083e7d4583b715160b5a4

app/views/images/_image.html.erb
kkajla12: Resolved in c09721214e194bb0af896166d78eeb6be2f3cdbb

test/controllers/images_controller_test.rb
kkajla12: Resolved in b1c72b783812136c166f548fac56e97045f884bf

app/views/images/share_form.html.erb
kkajla12: Resolved in ff177297b2d27ea37d2bd19851c553e7dbbc15da and 59ea6f068830283e21363d3bd92631b271120e08

app/views/images_mailer/share_image_email.html.erb
kkajla12: Resolved in 06926758bd23f86a056af3fe6e238b6592094209 and 027a20be359da2cb4a648c447fb05b6b2aad8787

app/views/images_mailer/share_image_email.html.erb
```diff
@@ -0,0 +1,9 @@
...
+<% end %>
+<br>
+<p>If you haven't already, try out <%= link_to 'Image Sharer', root_url, id: 'email-link' %> for yourself!</p>
```
kkajla12: Resolved in 06926758bd23f86a056af3fe6e238b6592094209 and 027a20be359da2cb4a648c447fb05b6b2aad8787

test/controllers/images_controller_test.rb
```diff
@@ -138,4 +138,102 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_redirected_to images_path
     assert_equal 'Image deleted successfully!', flash[:success]
   end
+
+  test 'should get share image form' do
+    image = Image.create!(url: 'http://images/0', tag_list: %w(0 1 2))
```
rmacklin: It's not related to this PR, but I'm surprised this URL passes validation. It seems the gem you are using is buggy (it also allows `http://example..com`)

kkajla12: Hmm, it's interesting that it allows ```http://example..com```. This is the gem I used: https://github.com/perfectline/validates_url

rmacklin: Yeah I know. I was just pointing out that it seems to have some bugs (I discovered the `example..com` one after seeing that it allowed this URL with no TLD). As mentioned, this is unrelated to the share image PR.

bboe: @kkajla12 this might be a good opportunity to contribute to an open source project. Perhaps you can Ryan can pair on a pull request that fixes some of those issues in the gem.

Regardless of fixing the gem or not, it would seem that this one example is a bug, and should e fixed. Can you make an issue for it?

kkajla12: @bboe Ryan and I actually discussed this yesterday after Richard pointed out the bug, and we definitely think it would be a good idea to open an issue for it and submit a fix.

kkajla12: @bboe there actually is already an [issue](https://github.com/perfectline/validates_url/issues/45) for this case, but I don't see a PR associated with it, so we can create one.

app/views/images_mailer/share_image_email.text.erb
```diff
@@ -0,0 +1,11 @@
+Cool! Your friend shared this image with you:
+
+<%= @image.url %>
+=========================================================
+
+Your friend said:
+<%= @message %>
```
rmacklin: What happens if they didn't provide a message?

test/mailers/images_mailer_test.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def html_email
+    <<HTML
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\" />
+    <style>
+      /* Email styles need to be inline */
```
rmacklin: FWIW, you can remove this part from your layout since you don't have any styles right now

app/views/images/share_form.html.erb
```diff
@@ -0,0 +1,27 @@
+<h1 class="flex-center">Share This Image</h1>
+<div class="flex-center">
+  <div class="tile js-image">
+    <%= link_to(image_tag(@image.url, class: 'tile__image'), image_path(@image), class: 'js-image-link') %>
+    <div class="tile__tags js-tags text-xs-center">
+      <% @image.tag_list.each do |tag| %>
+          <%= link_to(tag, images_path(tag_filter: tag), class: 'chip') %>
+      <% end %>
```
rmacklin: I don't think it's necessary to include the tags on the share page, as they won't be included in the email.

app/views/images/share_form.html.erb
```diff
@@ -0,0 +1,27 @@
...
+<div class="row flex-center">
+  <div class="col-md-6">
+    <%= simple_form_for @form, url: share_image_path(@image), html: { novalidate: true } do |f| %>
+        <%= f.input :recipient,
+                    label: 'Send To',
+                    error_html: { class: 'text-danger', id: 'recipient_error' } %>
+        <%= f.input :message,
+                    as: :text,
+                    label: 'Message (Optional)',
+                    required: false,
```
rmacklin: Since there is no presence validation on the `message` attribute, I don't think this is necessary

config/routes.rb
```diff
@@ -1,4 +1,8 @@
 Rails.application.routes.draw do
   root 'images#index'
-  resources :images, only: [:index, :new, :create, :show, :destroy]
+  resources :images, only: [:index, :new, :create, :show, :destroy] do
+    member do
+      get 'share_form'
```
rmacklin: This can be a symbol like the other actions.

config/routes.rb
```diff
@@ -3,6 +3,7 @@
   resources :images, only: [:index, :new, :create, :show, :destroy] do
     member do
       get 'share_form'
+      post 'share'
```
rmacklin: This can be a symbol like the other actions.

test/controllers/images_controller_test.rb
```diff
@@ -200,4 +200,40 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+        recipient: '',
+        message: 'Hello, World!'
+      })
+    end
+
+    assert_response :unprocessable_entity
+    assert_select '#recipient_error', "can't be blank"
+  end
+
+  test 'should not allow share with invalid recipient' do
```
rmacklin: This test and the previous one should be added in 9984ffffef04d7d41d1eb39536595b7ed065f312 rather than this commit (2dad03388459e32fd9dc0f4aaf31d96860817fda)

app/views/images/_image.html.erb
```diff
@@ -5,11 +5,16 @@
...
   </div>
-  <div class="tile__buttons text-xs-center">
-    <%= button_to 'Delete',
-                  { action: 'destroy', id: image.id },
-                  method: :delete,
-                  data: { confirm: 'Are you sure you want to delete this image?' },
-                  class: 'btn btn-danger js-delete-image-button' %>
+  <div class="tile__buttons row">
+    <div class="col-md-6 text-xs-right">
+      <%= link_to 'Delete',
```
rmacklin: Changing this shouldn't be grouped in with 2dad03388459e32fd9dc0f4aaf31d96860817fda whose commit message is:
> add share button to image cards, controller tests dependent on the views, and flow tests for sharing an image

test/flows/images_share_test.rb
```diff
@@ -0,0 +1,51 @@
...
+    cute_puppy_url = 'http://ghk.h-cdn.co/assets/16/09/980x490/landscape-1457107485-gettyimages-512366437.jpg'
+    Image.create!(url: cute_puppy_url, tag_list: 'puppy, cute')
+
+    images_index_page = PageObjects::Images::IndexPage.visit
+    image_to_share = images_index_page.images.find do |image|
+      image.url == cute_puppy_url
+    end
+
+    share_image_page = image_to_share.share!
+    Image.find(1).destroy!
```
rmacklin: This assumes the `id` of the image created on line 36 was `1`, which is not guaranteed. It's better to just save a reference to that image and call `destroy!` on that reference here.

test/flows/images_share_test.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    share_image_page = image_to_share.share!
+    message = 'Hello, World!'
+    share_image_page = share_image_page
+      .share_image!(recipient: 'invalid', message: message).as_a(PageObjects::Images::SharePage)
+    assert_equal 'is invalid', share_image_page.recipient.error_message
+    assert_equal 'invalid', share_image_page.recipient.value
+    assert_equal message, share_image_page.message.text
+
+    recipient = 'test@test.com'
```
rmacklin: Just curious, what's the purpose of this variable?

test/mailers/images_mailer_test.rb
rmacklin: This change can be deferred

app/views/images_mailer/share_image_email.text.erb
kkajla12: Whoops, forgot about the text email. Resolved in 2dad03388459e32fd9dc0f4aaf31d96860817fda

app/views/images/share_form.html.erb
kkajla12: Resolved in e886772335ffc0dae57bd11712beea28ade69ef2

app/views/images/share_form.html.erb
kkajla12: You're right. Resolved in afab86243236c1c0f3cd92bc3051b5868f9b62ef

config/routes.rb
kkajla12: Resolved in 9f39a7b36b4c89ed31ac3b5a653b4c9d2e5a88bc

config/routes.rb
kkajla12: Cool. Resolved in 9ceda8819b791254756bbd553e5f3228f6b08c57

test/flows/images_share_test.rb
kkajla12: It's not necessary. I was probably intending to use recipient again when first writing this test, but I ended up not doing so. Didn't catch that I could've just used the literal and gotten rid of the variable. Resolved in 53efad2ebae97d51ad9b9833198b75dc83d44f2e

test/flows/images_share_test.rb
kkajla12: Good catch. Resolved in cb05dd9dcf8e20d3394a0c386b398c900e3e0a4b

app/views/images_mailer/share_image_email.text.erb
```diff
@@ -3,8 +3,10 @@ Cool! Your friend shared this image with you:
 <%= @image.url %>
 =========================================================
 
-Your friend said:
-<%= @message %>
+<% if @message.present? %>
+  Your friend said:
+  <%= @message %>
+<% end %>
```
rmacklin: What else should accompany this change?

kkajla12: A test case to check that it's actually doing what it's supposed to

app/views/images_mailer/share_image_email.text.erb
rmacklin: See https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/48#discussion_r81006415

app/views/images_mailer/share_image_email.text.erb
kkajla12: Resolved in ee24510167e97c77fd8549e7d58d51769b113449

test/controllers/images_controller_test.rb
kkajla12: See 2ebf95bab2a05a33db9d50a2af44367002c9cd6e

app/views/images/_image.html.erb
kkajla12: See 3a13a9cf3dd57cc5c873ec149dd761115d424169

PR #49, opened 2016-09-20
Ropes configure send grid

kkajla12:
Configure emails to be sent using SendGrid in the production environment.


rmacklin:
It seems that this PR depends on #48. But that's backwards: the acceptance criteria for #48 depend on production email delivery being configured (while the ability to send emails doesn't depend on the ability to share images).

kkajla12:
@rmacklin, you're right. In order to configure / test that I had configured SendGrid properly, I wanted to have the email functionality from #48. My intention was to rebase this branch off of master once I had setup the configuration correctly, merge it in after review, and then rebase #48 off of the updated master.

rmacklin:
You can verify the ability to send emails just by firing up the Rails console to send a dummy email, and testing it in that direct and isolated manner is preferable: By not having the share image changes on the branch during the test, you're guaranteed that those changes didn't in some way affect your test of the sendgrid configuration.

rmacklin:
It looks like you manually added SendGrid to your review app. We want review apps to automatically get the SendGrid addon, so that we don't have to add it for each app. You can do this by specifying the addon in your app.json manifest.

config/environments/production.rb
```diff
@@ -57,6 +57,18 @@
   # config.active_job.queue_name_prefix = "base_#{Rails.env}"
   config.action_mailer.perform_caching = false
 
+  config.action_mailer.delivery_method = :smtp
+  config.action_mailer.default_options = { from: 'donotreply@appfolio.com' }
```
rmacklin: See https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/48#discussion_r79743469

kkajla12: I don't need to add it in this PR anyways. Fixing it in #48 will address the larger issue. Here I just removed it. Resolved in 4758e40e5e2281e84eff73ec1b2d73c68fb2cc95

kkajla12:
Added sendgrid heroku add-on in app.json in 827b6e33a574e25e0b6da13751e36ef2d3acf776

app.json
```diff
@@ -21,7 +21,8 @@
     }
   },
   "addons": [
-    "heroku-postgresql"
+    "heroku-postgresql",
+    "sendgrid"
```
rmacklin: It's safer to explicitly declare the free plans: `"heroku-postgresql:hobby-dev"` and `"sendgrid:starter"`. You can do this in its own commit

app.json
```diff
@@ -21,7 +21,8 @@
     }
```
rmacklin: You'll also want to declare the sendgrid environment variables to be inherited here.

app.json
kkajla12: The postgresql add-on is from the base project. It's not something that I added. I agree with declaring the tier explicitly to err on the side of safety, but in this case I think this works, since SendGrid's lowest tier is the free tier.

From [Heroku](https://devcenter.heroku.com/articles/setting-up-apps-using-the-heroku-platform-api#add-ons):

> When no tier is specified, the Platform API provisions the lowest tier, which for many add-ons is its free tier.

kkajla12: Resolved in 2c8c36c7afc8bfbf6845671da5e80b471d83432e

app.json
kkajla12: Resolved in 88a7b216a547bc179ff759bd6fe4da7705d2d4b7

PR #53, opened 2016-10-05
Ropes share image modal

### LINKED ISSUE #10: Share Image

kkajla12:
Closes #10


rmacklin:
Found this bug:

karan_modal_bug

This is a great opportunity to test-drive the bug fix. That is:

1. Write a test for this case which will fail with your current code
2. Fix the bug
3. Observe the the test now passes

app/views/images/show.html.erb
```diff
@@ -2,3 +2,7 @@
 <div class="flex-center">
   <%= render @image %>
 </div>
+
+<% content_for :javascript do %>
+  <%= javascript_include_tag 'images/index/index' %>
```
rmacklin: It's not appropriate for the show page to include the index page's javascript. It should have it's own file.

app/controllers/images_controller.rb
```diff
@@ -11,6 +11,8 @@ def index
     else
       @images = Image.order(created_at: :desc)
     end
+
+    @form = ShareImageForm.new
```
rmacklin: `@form` is a vague variable name. But this variable could be replaced with passing `ShareImageForm.new` directly in its place.

app/controllers/images_controller.rb
```diff
@@ -77,4 +75,40 @@ def image_params
...
+      format.json { render_json form_errors: true, status: :unprocessable_entity }
+      format.html { render :share_form, status: :unprocessable_entity }
+    end
+  end
+
+  def render_json(form_errors:, status:)
+    json = { form: render_to_string(partial: 'images/share_form',
+                                    formats: :html,
+                                    locals: { remote: true, image: @image }) }
+    json[:flash] = render_to_string(partial: 'layouts/flash_messages', formats: :html) unless form_errors
```
rmacklin: This `render_json` method is quite strange. It's as if the method is intended to be generic, but at the same time, it's coupled to the share form. I think everything would be a lot clearer if you just used specific methods for exactly what you need rather than trying to have this one `render_json` method try to handle lots of different things.

test/controllers/images_controller_test.rb
```diff
@@ -10,7 +10,7 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
 
     assert_response :ok
     assert_select 'h1', 'Images'
-    assert_select 'img' do |images|
+    assert_select '.js-image' do |images|
```
rmacklin: Changing `img` to `.js-image` everywhereshould be its own independent commit.

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,41 @@
...
+
+    this.$modal.on('hide.bs.modal', () => {
+      this.$modal.find('.js-image-modal-share-form')[0].reset();
+    });
+
+    this.$modal.on('ajax:success', (_event, data) => {
+      this.$modal.modal('hide');
+      $('#flash_messages').html(data.flash);
+      $('html, body').animate({ scrollTop : 0 }, 800);
+      this.$modal.find('.js-image-modal-share-form').html(data.form);
```
kkajla12: @rmacklin, this line is intended to reset the form after a successful submission.

kkajla12:
I fixed the bug in da19fcf61d530deb788390a7b98383c9e3097dc7 with the corresponding test coming in 85dec078f37d6b8ac92c739760bda724081e457c along with the rest of the flow tests.

app/assets/javascripts/images/share_modal.js
kkajla12: However, with the commits mentioned in https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/53#issuecomment-253914360 that were added to address the bug you pointed out above, this line might not be necessary anymore.

test/controllers/images_controller_test.rb
kkajla12: I'll go ahead and pull these changes out into their own commit when rebasing before merging.

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,41 @@
...
+    });
+
+    this.$modal.on('ajax:error', (_event, xhr) => {
+      if (xhr.status === 404) {
+        window.location.replace('/');
+      } else {
+        this.$modal.find('.js-image-modal-share-form').html(xhr.responseJSON.form);
```
kkajla12: Resolved in 24230cb73dc439e1b9fb8cfd0713fbe52c60d476

app/controllers/images_controller.rb
kkajla12: @rmacklin, I would like to keep this the way it is rather than specifying ```ShareImageForm.new``` within the view itself because I would need to change a few other things to make that work, and I like that I know what instance variables are available in the view from the controller. Would it be ok to change ```@form``` to something more specific like ```@share_image_form```?

app/assets/javascripts/images/share_modal.js
```diff
@@ -25,8 +25,11 @@ class ShareModal {
     this.$modal.on('ajax:error', (_event, xhr) => {
       if (xhr.status === 404) {
         window.location.replace('/');
-      } else {
+      } else if (xhr.status === 422) {
         this.$modal.find('.js-image-modal-share-form').html(xhr.responseJSON.form);
+      } else {
+        this.$modal.modal('hide');
+        alert('We were unable to share your the image at this time. Please try again later.');
```
rmacklin: UX comment: This is potentially a transient server error, and if you hide the modal, they'd have to re-enter what they already typed for the recipient and message. In light of this, I think it's better to just alert `'We were unable to share your the image at this time. Please try again later.'` without hiding the modal.

app/assets/javascripts/images/share_modal.js
```diff
@@ -25,8 +25,11 @@ class ShareModal {
     this.$modal.on('ajax:error', (_event, xhr) => {
       if (xhr.status === 404) {
         window.location.replace('/');
-      } else {
+      } else if (xhr.status === 422) {
```
rmacklin: One thing you could do to slightly improve readability is define constants at the top of the module that use the "human readable" names for these status codes:
```js
const NOT_FOUND = 404;
const UNPROCESSABLE_ENTITY = 422; 
```

app/assets/javascripts/images/share_modal.js
```diff
@@ -35,6 +35,8 @@ class ShareModal {
     this.$modal.find('.js-image-modal-image-preview').attr('src', modalImageUrl);
     this.$modal.find('.js-image-modal-share-form').attr('action', modalShareUrl);
     this.$modal.find('.js-image-modal-image-link').attr('href', modalShowUrl);
+    this.emptyForm = this.$modal.find('.js-image-modal-share-form')[0].outerHTML;
```
rmacklin: First, is it necessary to set `this.emptyForm` inside `setupModal`, which is called every time the modal is shown?

rmacklin: Second, your logic on lines 38 and 15 is quite seriously flawed, although you are not able to witness it in this particular case because you happen to be producing invalid HTML that the browser is automatically correcting, which counteracts the buggy logic. <sup>1</sup> Let me explain.

On this line (38), you're setting `this.emptyForm` to the _outerHTML_ of the `.js-image-modal-share-form` element (i.e. the `<form>` element). On line 15, you're using [jQuery's `html`](http://api.jquery.com/html/) method to set the _innerHTML_ of this element to `this.emptyForm`. What that means is, you're code is creating this DOM structure:
```html
<form>
  <form>
   </form>
</form>
```
However, in HTML a `<form>` element cannot contain another `<form>`. So when you call `html(this.emptyForm)`, the browser is removing the nested form structure, so you don't end up seeing this. But for the sake of demonstration, change `this.$modal.find('.js-image-modal-share-form')` to `this.$modal.find('.js-image-modal-share-form').parent()` on lines 38 and 15, so that you're selecting the parent `<div class="col-md-11">` element, rather than the form. If you inspect the DOM, you'll see that _each time_ you hide the modal, you nest another `<div class="col-md-11">` inside the previous one, quickly ending up with:
```html
<div class="col-md-11"> 
  <div class="col-md-11">
    <div class="col-md-11">
      <div class="col-md-11">
        <div class="col-md-11">
          .
            .
              .
                <div class="col-md-11">
                  <div class="col-md-11">
                    <div class="col-md-11">
                      <form>
                      </form>
                    </div>
                  </div>
                </div>
              .
            .
          .
        </div>
      </div>
    </div>
  </div>
</div>
```
**D I V C E P T I O N**

Leo squinting

In many cases, this divception won't appear to affect the page (because by default divs are just containers), but if the nesting gets large enough you could cause noticeable slowdown. Regardless, it's _always_ unnecessary and should be avoided.

<sup><sup>1</sup> I'd argue that while it seems like you're getting "lucky", you're actually getting really _unlucky_ because this rare case of improperly setting the element's _innerHTML_ to it's _outerHTML_ happens to "work", so it appears like the logic isn't flawed -- which means you may copy it somewhere else where it doesn't produce invalid HTML and thus does not get countered by the browser, and you end up nesting elements instead of replacing them.</sup>

app/assets/javascripts/images/share_modal.js
rmacklin: Right, this line isn't necessary if you always reset to an empty form. That means it's also not necessary for the controller to send `data.form`

rmacklin:
https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/53#discussion_r82704811 appears as under an outdated diff for some reason, but it has not been addressed.

app/controllers/images_controller.rb
kkajla12: @rmacklin, I don't think I understand this. If I directly set the first argument of `simple_form_for` to `ShareImageForm.new`, anytime I render the form it will render as an empty form. Even if the form was just submitted with errors. Could you please explain?

rmacklin: Well first you'd refactor `simple_form_for @form` to `simple_form_for share_form`. Generally partials shouldn't depend on instance variables because partials (unlike templates) aren't tied to any particular controller action. By changing it to a local instead of an instance variable, you no longer have this issue. You'd change
```erb
<%= render partial: 'share_form', locals: { remote: true, image: nil } %>
```
to
```erb
<%= render partial: 'share_form', object: share_form, locals: { remote: true, image: image } %>
```
in the share_image_modal partial. To render the modal in the show.html.erb template, you'd use
```erb
<%= render partial: 'share_image_modal', locals: { form: ShareMailForm.new, image: @image } %>
```
(for show.html.erb, you should _not_ pass `nil` for the image like you are currently doing, because we know the image that can be shared from the show page)

And to render the modal in index.html.erb, you'd use:
```erb
<%= render partial: 'share_image_modal', locals: { form: ShareMailForm.new, image: nil } %>
```
(for index.html.erb, we don't know which image will be shared)

kkajla12: I just found out about the instance variable by the same name as the partial being available implicitly. Thanks for the explanation!

rmacklin: To clarify, what I showed above is a _local_ (not _instance_) variable with the same name as the partial, which is set to the object I (explicitly) passed via the `object` parameter.

See also: http://guides.rubyonrails.org/layouts_and_rendering.html#using-partials

kkajla12: Yeah, sorry, I used some poor wording in my previous comment. I learned of the feature through the guide you just referenced. Thanks for clarifying

app/views/images/_share_image_modal.html.erb
```diff
@@ -0,0 +1,26 @@
+<div class="modal fade" id="share_image_modal" tabindex="-1" role="dialog" aria-labelledby="share_image_model_label" aria-hidden="true" style="display: none">
```
kkajla12: Resolved in 550a4ce34a8c3bf9b2883c4a610abfbf62107840

app/controllers/images_controller.rb
kkajla12: Refactored in df6fd3c0a6b1e16df4d35bcef138ccc7a48c76af

app/assets/javascripts/images/share_modal.js
kkajla12: Because of my misunderstanding of innerHTML and outerHTML (which led me to use outerHTML), addressed by https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/53#discussion_r83593741, I thought that the empty form needed to be saved each time the modal was shown because the form action url could change each time depending on which image the user chose to share. After having refactored the JS to only use innerHTML, this is no longer true, so I can move this into the modal's constructor. This is resolved in b83fcbd1343fcb5f42e306058713853eeee94fd3.

kkajla12: @rmacklin, thank you for the thorough explanation. I've switched to only assigning innerHTML to innerHTML in b83fcbd1343fcb5f42e306058713853eeee94fd3.

app/assets/javascripts/images/share_modal.js
kkajla12: Resolved in b83fcbd1343fcb5f42e306058713853eeee94fd3

app/controllers/images_controller.rb
kkajla12: Refactored in 4c9fd1136093bef1707e3042d41ac7bdc8909a7b

app/assets/javascripts/images/share_modal.js
kkajla12: Resolved, following a slight refactor, in 21f3a65a36d3372c585b602a29b9e57c0c7da39f

app/views/images/show.html.erb
kkajla12: Resolved in 7e3c6458b2bde1aa1550d6cc7332123ea4aa2422

test/flows/images_share_modal_test.rb
```diff
@@ -68,4 +68,26 @@ class ImagesShareModalTest < FlowTestCase
...
+
+    image_to_share.share_using_modal do |modal|
+      modal.share_image(recipient: 'invalid', message: 'Hello, World!')
+      assert_equal 'is invalid', modal.recipient.error_message
+      modal.close
+    end
+
+    image_to_share.share_using_modal do |modal|
+      assert_empty modal.recipient.value
+      assert_empty modal.message.value
```
kkajla12: Resolved in 30617697b3dfc040301583e7ec29ec0a02d0b36f

config/initializers/assets.rb
```diff
@@ -9,3 +9,4 @@
 # application.js, application.scss, and all non-JS/CSS in app/assets folder are already added.
 Rails.application.config.assets.precompile += %w( mdb.js )
 Rails.application.config.assets.precompile += %w( images/index/index.js )
+Rails.application.config.assets.precompile += %w( images/show/index.js )
```
rmacklin: These should be combined:
```rb
Rails.application.config.assets.precompile += %w(
  images/index/index.js
  images/show/index.js
)
```

app/controllers/images_controller.rb
```diff
@@ -76,39 +76,46 @@ def image_share_form_params
...
     end
   end
 
-  def render_json(form_errors:, status:)
-    json = { form: render_to_string(partial: 'images/share_form',
-                                    formats: :html,
-                                    locals: { remote: true, image: @image }) }
-    json[:flash] = render_to_string(partial: 'layouts/flash_messages', formats: :html) unless form_errors
-    render json: json, status: status
+  def render_form_to_string
```
rmacklin: This is hardcoded to render the _share_ form, so the name should reflect that: `render_share_form_to_string`

app/controllers/images_controller.rb
```diff
@@ -76,39 +76,46 @@ def image_share_form_params
...
+  def respond_to_invalid_image
     @form = ShareImageForm.new
-    flash[:error] = message
+    flash[:error] = 'Oops! The image you are trying to share was not found!'
 
     respond_to do |format|
-      format.json { render_json form_errors: false, status: :not_found }
+      format.json do
+        render json: {
+          form: render_form_to_string
```
rmacklin: It's not clear why you're rendering the form in this case.

app/assets/javascripts/images/share_modal.js
```diff
@@ -19,16 +23,15 @@ class ShareModal {
       this.$modal.modal('hide');
       $('#flash_messages').html(data.flash);
       $('html, body').animate({ scrollTop : 0 }, 800);
-      this.$modal.find('.js-image-modal-share-form').html(data.form);
+      this.$modal.find('.js-image-modal-share-form').html(this.emptyForm);
```
rmacklin: Is this line necessary?

app/assets/javascripts/images/share_modal.js
```diff
@@ -1,8 +1,12 @@
 import $ from 'jquery';
 
+const NOT_FOUND = 404;
+const UNPROCESSABLE_ENTITY = 422;
+
 class ShareModal {
   constructor(modalSelector) {
     this.$modal = $(modalSelector);
+    this.emptyForm = this.$modal.find('.js-image-modal-share-form')[0].innerHTML;
```
rmacklin: Just so you know, this is equivalent to just calling `this.$modal.find('.js-image-modal-share-form').html()`

config/initializers/assets.rb
```diff
@@ -9,3 +9,4 @@
 # application.js, application.scss, and all non-JS/CSS in app/assets folder are already added.
 Rails.application.config.assets.precompile += %w( mdb.js )
```
kkajla12: Resolved in 5e1e962abc63a0f227a3f281eff723356063d6ec in #60.

app/controllers/images_controller.rb
kkajla12: It's not needed. Thanks for pointing that out. Resolved in 9fb7f79f219bca476505cfff8dda2408c66aac06

app/controllers/images_controller.rb
kkajla12: Resolved in 9fb7f79f219bca476505cfff8dda2408c66aac06

app/controllers/images_controller.rb
```diff
@@ -102,14 +98,14 @@ def respond_to_share_image_error(form:)
     respond_to do |format|
       format.json do
         render json: {
-          form: render_form_to_string(form: form)
+          form: render_share_form_to_string(form: form)
```
rmacklin: If you take a look at the invocation of the `render_share_form_to_string` method here, using a keyword argument for `form` makes it a bit overly redundant/verbose. Since there is only one argument, and the method name describes what that argument is, it's probably cleaner to just use a regular positional parameter:
```rb
form: render_share_form_to_string(form)
```

test/flows/images_share_modal_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    cute_puppy_image.destroy!
+
+    image_to_share.share_using_modal do |modal|
+      modal.share_image(recipient: 'test@test.com', message: 'Hello, World!')
+    end
+
+    assert_equal 'Oops! The image you are trying to share was not found!',
+                 images_index_page.flash_message(:error)
```
rmacklin: This `images_index_page` is a stale reference because the client reloaded the page, which means this line is flaky. You'll want to use
```ruby
images_index_page =
  AePageObjects.browser.current_window.change_to PageObjects::Images::IndexPage
```
to get a reference to the new (not stale) page

test/flows/images_share_modal_test.rb
```diff
@@ -0,0 +1,94 @@
...
+      modal.message.set('Hello, World!')
+      modal.close
+    end
+
+    image_to_share.share_using_modal do |modal|
+      assert_empty modal.recipient.value
+      assert_empty modal.message.value
+      modal.close
+    end
+  end
```
rmacklin: In this test and the one after it, the second share modal should be for a different image than the first one, so that you can also assert that the image url is properly updated when the second modal is opened.

test/flows/images_share_test.rb
```diff
@@ -43,8 +60,9 @@ class ImagesShareTest < FlowTestCase
     cute_puppy_image.destroy!
 
     images_index_page = share_image_page
-      .share_image!(recipient: 'test@test.com', message: 'Hello, World!').as_a(PageObjects::Images::IndexPage)
-    assert_equal 'Oops! The image you are attempting to share was not found!',
+      .share_image!(recipient: 'test@test.com', message: 'Hello, World!')
+      .as_a(PageObjects::Images::IndexPage)
```
rmacklin: `as_a` is only necessary when switching to a page other than the first one in the list in `window.change_to`, so if you change the order there as explained in my other comment, you can remove this `as_a` cast

test/page_objects/images/image_card.rb
```diff
@@ -28,7 +28,15 @@ def delete_and_confirm!
 
       def share!
         node.click_link('Share')
-        window.change_to(SharePage)
+        window.change_to(SharePage, IndexPage)
+      end
+
+      def share_using_modal
```
rmacklin: This should be called just `share`, and the other method, which is the less likely case (only applies with JS disabled), should be called something like `open_share_page!`

test/flows/images_share_modal_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    image_to_share.share_using_modal do |modal|
+      assert_empty modal.recipient.value
+      assert_empty modal.message.value
+      assert_empty modal.recipient.error_message
+      modal.close
+    end
+  end
+end
```
rmacklin: There are no tests that test sharing an image from the show page, so we're not testing that the share modal javascript is included on that page.

test/page_objects/images/share_image_modal.rb
```diff
@@ -0,0 +1,24 @@
+module PageObjects
+  module Images
+    class ShareImageModal < AePageObjects::Element
+      form_for :share_image_form do
+        element :recipient, locator: '#share_image_form_recipient'
+        element :message, locator: '#share_image_form_message'
```
kkajla12: haha, nope.

test/page_objects/images/share_image_modal.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      def image_url
+        node.find('img')[:src]
+      end
+
+      def share_image(recipient: nil, message: nil)
+        share_image_form.recipient.set(recipient) unless recipient.nil?
+        share_image_form.message.set(message) unless message.nil?
+        node.click_button('Share Image')
+      end
```
rmacklin: This page object has some duplication with the `SharePage`. Peter came up with a good way to DRY the page objects, which you could ask him about.

rmacklin:
Ok, I left just one minor comment on the prod code: https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/53#discussion_r84777551
You can rebase to squash all the fixups, address the history-altering comments, and fix the merge conflict. (And you can fixup that one comment at the same time.)

My other comments are about the tests. Personally, I would do the initial rebase and then address these comments in subsequent fixup commits, but if you'd rather address them mid-rebase, you can do so.

app/controllers/images_controller.rb
kkajla12: I actually thought about this while looking the code over earlier, but I didn't change it. I agree with you. Resolved in 454dd8319818a60f4c592cbca1986c0b83521977

test/page_objects/images/image_card.rb
```diff
@@ -28,7 +28,15 @@ def delete_and_confirm!
 
       def share!
         node.click_link('Share')
-        window.change_to(SharePage)
+        window.change_to(SharePage, IndexPage)
```
kkajla12: @rmacklin, I think it might be the other way around. I tried switching it and removing the explicit cast using `as_a` and it didn't work. I switched the order back, keeping the explicit cast out, and that worked.

kkajla12: The documentation says what you said, but that's not working for me.

rmacklin: Sorry, I misread this diff. `share!` here (i.e. `open_share_page!` per https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/53#discussion_r84782835), should only return `SharePage`. I had been mistaking this line for https://github.com/AppFolioOnboarding/image-sharer-kkajla12/blob/591fea570bd3f47afc438cc7980f5da10965e21a/test/page_objects/images/share_page.rb#L16, where the form is submitted, and where the order should indeed be `IndexPage, SharePage`.

app/controllers/images_controller.rb
kkajla12: This was addressed before rebase: https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/53/files#diff-a7ae6992fe889ec0c720925f4275f636R108

test/page_objects/images/image_card.rb
kkajla12: Resolved in f8502cf52fc529da414df0baa27148c54bc31673

test/page_objects/images/image_card.rb
kkajla12: Resolved in f8502cf52fc529da414df0baa27148c54bc31673

test/page_objects/images/share_image_modal.rb
kkajla12: Resolved in f8502cf52fc529da414df0baa27148c54bc31673

test/flows/images_share_modal_test.rb
```diff
@@ -39,30 +39,40 @@ class ImagesShareModalTest < FlowTestCase
...
 
-    image_to_share.share_using_modal do |modal|
+    image_to_share.share do |modal|
       modal.recipient.set('test@test.com')
       modal.message.set('Hello, World!')
       modal.close
     end
 
-    image_to_share.share_using_modal do |modal|
+    image_to_share = images_index_page.images.find do |image|
```
rmacklin: This test (and the subsequent one) would be clearer if you didn't use `image_to_share` for the variable names, but rather distinct names for the different images (e.g. `cute_puppy_image`, `ugly_cat_image`).

test/flows/images_share_modal_test.rb
rmacklin: That commit now opens the modal for different images, which is good, but you didn't add the assertions about the image URL in the modal corresponding to the image we expect.

Additionally: https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/53#discussion_r84958035

test/flows/images_share_modal_test.rb
```diff
@@ -71,20 +81,28 @@ class ImagesShareModalTest < FlowTestCase
...
       assert_equal 'is invalid', modal.recipient.error_message
       modal.close
     end
 
-    image_to_share.share_using_modal do |modal|
+    image_to_share = images_index_page.images.find do |image|
+      image.url == ugly_cat_url
+    end
+
+    image_to_share.share do |modal|
```
rmacklin: See https://github.com/AppFolioOnboarding/image-sharer-kkajla12/pull/53#discussion_r84958138

test/flows/images_share_modal_test.rb
kkajla12: Resolved in f99be4706a5663ddbaf40e150a8e59e803b484ac

test/flows/images_share_modal_test.rb
kkajla12: Resolved in f99be4706a5663ddbaf40e150a8e59e803b484ac

test/page_objects/images/share_image_modal.rb
kkajla12: Refactored in 171a04871c985a446033572c4e2bdc528a18d7ee

PR #54, opened 2016-10-05
Override Bootstrap Classes and Modify Style to Render Application Consistently on Firefox

kkajla12:
This PR addresses an inconsistency in the way that Chrome and Firefox handle element positioning using the height attribute and flex-box. After adding the sticky footer in #31, Chrome was rendering the application as intended, but Firefox was rendering pieces of the content in a different ordering vertically. It was determined that MDBootstrap's ```flex-center``` class (used to center elements on a page) was applying ```height: 100%``` which caused vertical overflow in Firefox and rendered elements with incorrect ordering, often far below where the viewport was currently positioned. This PR overrides flex-center to not set the height attribute automatically rather than apply ```height: 100%``` to fix this issue.


app/assets/stylesheets/bootstrap_overrides.scss
```diff
@@ -0,0 +1,3 @@
+.flex-center {
```
rmacklin: From your PR description:
> It was determined that MDBootstrap's flex-center class ...

It sounds like this is really `mdb_overrides`, not `bootstrap_overrides`, unless that class is in fact from bootstrap, not mdb.

Also, overriding library styles to fix browser bugs is a good example of where code comments are valuable.

PR #55, opened 2016-10-05
rename bootstrap_overrides.scss to mdb_overrides.scss and add a comme…

kkajla12:
This PR addresses comments left post merge on #54.


kkajla12:
@rmacklin, I did notice that git classified it as a delete followed by the creation of a new file after I modified the renamed file, but like you, I figured it would be ok in this case. I'll keep it in mind in the future.

PR #58, opened 2016-10-10
Ropes required tag association

### LINKED ISSUE #12: Required Tag Association
#### As a user I want all images to have at least one tag.

__Story__:

The tag feature has been a hit. Our users love being able to filter images by
cats, SR-71s, and "AppFolio tshirts" to name a few. In fact, we have discovered
that images without any tags receive significantly less views than those with
tags. As a result, we have decided to require all images to have at least one
tag.

__Acceptance criteria__:
- [x] I cannot save an image without adding at least one tag.
- [x] An error message is associated with the appropriate input field when a
tag is not provided.
- [x] All images have at least one tag.

__Discussion Topic__:
- [x] What do you do about existing images that have no tags?

__Dependencies__:
- Image Tags

kkajla12:
Closes #12


db/migrate/20161007163356_add_tag_to_images_without_tags.rb
```diff
@@ -0,0 +1,35 @@
+class AddTagToImagesWithoutTags < ActiveRecord::Migration[5.0]
+  def change
+    ActiveRecord::Base.connection.execute("INSERT INTO Tags (name) VALUES('');")
```
kkajla12: @pololee, this line just creates a tag with empty string as the value, and this is the tag I then add to all of the existing images without a tag. I decided to do this because users are not able to create images with empty string as a tag. This way I can ensure that the tag I am applying to all of the existing images is not one that users can add themselves too.

db/migrate/20161007163356_add_tag_to_images_without_tags.rb
```diff
@@ -0,0 +1,35 @@
+class AddTagToImagesWithoutTags < ActiveRecord::Migration[5.0]
+  def change
```
kkajla12: @pololee, I don't think Rails would know how to rollback this migration, however, I don't think this is a migration that can really be rolled back in a nice way. We don't really have any way of knowing which images didn't have a tag before. With that being said, I think it would be best to use `change` because if you use `up` then it must be accompanied by a corresponding `down` and we don't really know what `down` would consist of. Would you agree?

test/models/image_test.rb
```diff
@@ -2,25 +2,32 @@
 
 class ImageTest < ActiveSupport::TestCase
   test 'url is valid' do
-    image = Image.new(url: 'https://test.com')
+    image = Image.new(url: 'https://test.com', tag_list: %w(0 1 2))
```
kkajla12: Refactored in 0c7e092ac33eb53e8b535c3e02ba369f16773665. How's that?

db/migrate/20161007163356_add_tag_to_images_without_tags.rb
pololee: 👍

test/models/image_test.rb
```diff
@@ -1,10 +1,12 @@
 require 'test_helper'
 
 class ImageTest < ActiveSupport::TestCase
-  test 'url is valid' do
-    image = Image.new(url: 'https://test.com', tag_list: %w(0 1 2))
+  setup do
+    @image = Image.new(url: 'https://test.com', tag_list: %w(0 1 2))
```
pololee: I guess I wasn't clear about my comment. Here is my idea.

1. have a private method

  ```ruby
  private

  def default_params(url: 'https://test.com', tag_list: %w(0 1 2))
    {
      url: url,
      tag_list: tag_list
    }
  end
  ```

2. Then in each method, you just need to call this private method
e.g. 
"url must be present": `Image.new(default_params(url: ''))`
"tag_list must be present": `Image.new(default_params(tag_list: nil))`

3. In this way, when you have a new attribute or a simple validation, you can just change the default_params to return a valid params and you don't need to modify any test.

what do you think?

db/migrate/20161007163356_add_tag_to_images_without_tags.rb
kkajla12: Yes, but it would result in just one tag `''` being created because tags must be unique.

pololee: It just feels weird that we require at least one tag, but allow `''` to be a valid tag. But I am OK with keeping what you have here.

kkajla12: I agree that it's a little weird. However, it is only possible to create `''` as a tag manually through the database since the UI enforces that there must be something in `tag_list` for a valid form submission. Therefore, it's not possible for a user to create `''` as a tag, which is why I chose to make it the default tag added in this migration. That way there is a clear distinction between the user created tags and the default tag for the images that previously didn't have a tag.

pololee: 👍  
I remember I had a similar issue. We end up using `'n/a'`.

test/models/image_test.rb
kkajla12: It took me a little while, but I now see the benefit of this over creating an image in the setup block.

kkajla12: Resolved in 04775f953e2456dbb6045c7d8db9315b32c8108b

pololee: 👍

PR #59, opened 2016-10-11
Ropes modify tag list

### LINKED ISSUE #13: Modify Tag List
#### As a user I want to edit the tag list for an image.

__Story__:

Image tags are awesome -- our user's love them! Tags help us make sense of the
plethora of images added to our system. But how annoying is it when our users
make spelling mistakes in their image tags? Super annoying! How awesome, and
useful would it be to allow for tags to be edited after an image is added?
Super awesome, and useful!

__Acceptance criteria__:
- [x] There is a link to edit the tags for an image.
- [x] The image whose tags are to be modified is displayed on the edit page.
- [x] On the edit page, I can change the tags to any valid tag list.
- [x] After saving the edit page with valid tags, my changes are visible
  throughout the rest of the application.

__Discussion__:
- [x] Should an image's URL be editable?

__Dependencies__:
- Image Tags

kkajla12:
Closes #13


test/controllers/images_controller_test.rb
```diff
@@ -85,6 +85,76 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    image = Image.create!(url: 'http://images/0', tag_list: 'A, C')
+
+    patch image_path(image), params: {
+      image: {
+        tag_list: ''
+      }
+    }
+    assert_response :unprocessable_entity
+
+    assert_select '.image_tag_list span.text-danger', "can't be blank"
```
bboe: @weixing2014 FYI, `assert_template` was removed in rails 5 for a good reason. We don't really care about the implementation, but are more concerned with the result.

rmacklin: Asserting that the edit form is displayed (in addition to the specific assertion about the error message) would be a roughly equivalent replacement assertion for `assert_template :edit`

kkajla12: I updated this test like Richard said in 6e1c8c7354d56e9c71577577bfba72d9e419b712

PR #67, opened 2016-11-01
Ropes add test record creation helper methods

kkajla12:
Add helper methods in `TestCase` class which encapsulate calls to `create!` and `new` with default attribute values for existing ActiveRecord and ActiveModel objects in the project. This provides the benefit of only requiring attributes to be updated in one location if they are ever changed.


bboe:
👍

